PR #1, opened 2020-11-26
Initial Application Setup

schaefi:
I've setup keg as python project.

To start simply use tox as follows:

```
$ tox -e devel
$ . .tox/3/bin/activate
$ keg

[ INFO    ]: 16:44:12 | This is keg version 0.0.1
```

There is not more it can do so far. I added just the fundamental basics
which are version, logging and exception classes plus their tests.

Other targets I added are:

* Packaging

  ```
  make build
  ```

  Also manages the changelog setup from the git history.
  Source data is created from distutils, thus we could also
  submit to pip if we want. Find rpm data below dist/

* Unit Tests

  ```
  tox -e unit_py3_6
  tox -e unit_py3_8
  ```

  Depending on your host missing interpreters are skipped

* Python style check

  ```
  tox -e check
  ```

Ready to rumble :-) Let me know what you think

Thanks


schaefi:
Also enabled travis for running the unit tests

jgleissner (approved):

jgleissner:
Looks good to me, thanks!

PR #5, opened 2020-12-22
Initial code drop

jgleissner:
Initial code drop, supports generation of config.kiwi and config.sh.

### DIFF
--- a/keg/generator.py
+++ b/keg/generator.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020 SUSE Software Solutions Germany GmbH. All rights reserved.
+# Copyright (c) 2021 SUSE Software Solutions Germany GmbH. All rights reserved.
 #
 # This file is part of keg.
 #
--- a/keg/generator.py
+++ b/keg/generator.py
@@ -1,0 +1,82 @@
+# Copyright (c) 2020 SUSE Software Solutions Germany GmbH. All rights reserved.
+#
+# This file is part of keg.
+#
+# keg is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# keg is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with keg. If not, see <http://www.gnu.org/licenses/>
+#
+from jinja2 import Environment, FileSystemLoader
+import os
+
+from keg import (
+    image_definition,
+    kiwi_description,
+    template_functions
+)
+from keg.exceptions import KegError, KegKiwiValidationError
+
+
+def create_image_description(image_source,
+                             recipes_root,
+                             dest_dir,
+                             log,
+                             force=False):
+    """
+    Create a KIWI image description in dest_dir from image_source.
+
+    Expects a keg-recipes structure at recipes_root and image_source
+    referring to a directory under images in recipes_root.
+    """
+    # Sanity check image source location
+    if not os.path.exists(os.path.join(recipes_root, 'images', image_source)):
+        errmsg = 'Source directory for {} does not exist.'.format(image_source)
+        if force:
+            log.warning(errmsg)
+        else:
+            raise KegError(errmsg)
+
+    img = image_definition.ImageDefinition(image_source, recipes_root)
+    img.populate()
+
+    schemas_dir = os.path.join(recipes_root, 'schemas')
+    env = Environment(loader=FileSystemLoader(schemas_dir))
+    env.globals['keg_funcs'] = template_functions
+
+    kiwi_templ = env.get_template('{}.kiwi.templ'.format(img['schema']))
+    kiwi_doc = kiwi_templ.render(data=img.get_data())
+
+    outfile = os.path.join(dest_dir, 'config.kiwi')
+    if os.path.exists(outfile) and not force:
+        raise KegError('{} exists, use force to overwrite.'.format(outfile))
+    with open(outfile, 'w') as fd:
+        fd.write(kiwi_doc)
+
+    kiwi_desc = kiwi_description.KiwiDescription(outfile)
+    try:
+        kiwi_desc.validate_description()
+    except KegKiwiValidationError:
+        if force:
+            log.warning('KIWI description validation error')
+            # FIXME: find out how to get details
+            pass
+        else:
+            raise
+
+    config_templ = env.get_template('{}.config.sh.templ'.format(img['schema']))
+    config_sh = config_templ.render(data=img.get_data())
+
+    outfile = os.path.join(dest_dir, 'config.sh')
+    if os.path.exists(outfile) and not force:
+        raise KegError('{} exists, use force to overwrite.'.format(outfile))
+    with open(outfile, 'w') as fd:
+        fd.write(config_sh)
--- a/keg/generator.py
+++ b/keg/generator.py
@@ -21,7 +21,8 @@ import os
 from keg import (
     image_definition,
     kiwi_description,
-    template_functions
+    template_functions,
+    utils
 )
 from keg.exceptions import KegError, KegKiwiValidationError
 
@@ -73,8 +74,10 @@ def create_image_description(image_source,
         else:
             raise
 
+    script_roots = [ os.path.join(recipes_root, 'data') ] + data_roots
+    script_lib = utils.load_scripts('scripts', script_roots, img['include-paths'])
     config_templ = env.get_template('{}.config.sh.templ'.format(img['schema']))
-    config_sh = config_templ.render(data=img.get_data())
+    config_sh = config_templ.render(data=img.get_data(), scripts=script_lib)
 
     outfile = os.path.join(dest_dir, 'config.sh')
     if os.path.exists(outfile) and not force:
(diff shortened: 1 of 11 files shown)


schaefi (changes requested):
Thanks for starting the command line. Here is what I think we need to discuss:

1. The code has no tests. We agreed on full test coverage. That's also why the PR is in failed state. If you want me to help with the tests, let me know

2. I don't like argparser :) This is mostly personal taste but there are also some technical aspects that speaks for docopt: http://docopt.org/. First is readability. The interface of the command line is a description language. Also using docopt makes it easy for people to use keg for plugins which I can imagine will be one use case for other people using keg. Usage of the command line is expressed in code in the exact same way as if one calls "keg --help", documentation of the command line belongs to the code and therefore you get that part for free. Extending the commandline with additional features is way more sophisticated than with argparse. I agree the keg cmdline might be very simple but if not argparse becomes cumbersome which it already is to be honest with 50 lines to just express a few options.

As I said argument parsing is mostly personal taste so I'll go with the majority of the votes.
On the missing unit tests there should be not much of a discussion though ;)

Thanks

schaefi:
Example describing the command interface with docopt:

```python
"""
usage: keg -h | --help
       keg [options] --source=<source> --dest-dir=<dest_dir>

options:
    -r recipes_root, --recipes-root=<recipes_root>
        root of recipes

    -f, --force
        force run

    -v, --verbose
        be verbose
"""
from docopt import docopt

arguments = docopt(__doc__, version='1.0')

print(arguments)
```

schaefi:
* ```--version``` you get for free ```keg --version```

* ```keg -h``` or ```keg --help``` prints the usage as defined in the docstring.

* The parameter parsing ends in a dict as you can see when calling the above code

jgleissner:
Hi @schaefi, thanks for the holiday review. Ack on the unit tests. I wanted to push what I have so you can take a first look, it's not finished yet. I should have more verbose on this in the comment.

Re docopt, I haven't used it so far and I just used argparse because it seemed easy enough and suitable for a simple cli tool. I've used click before but it confused me. ;) I agree that docopt looks a lot neater though, so perhaps worth switching.

keg/generator.py
```diff
@@ -0,0 +1,82 @@
...
+def create_image_description(image_source,
+                             recipes_root,
+                             dest_dir,
+                             log,
+                             force=False):
+    """
+    Create a KIWI image description in dest_dir from image_source.
+
+    Expects a keg-recipes structure at recipes_root and image_source
+    referring to a directory under images in recipes_root.
```
rjschwei, line 38: I do not follow why the separation between `image_source` and `recipes_root` is necessary. If `image_source` is expected to be part of `recipes_root`. This need better clarification in the doc and/or code or should be merged into 1 argument. Also `recipes_root` cannot be singular, it has to be a list as we need to be able to build a description from multiple tree definitions by merging. To keep it simple I think we should take a `last one wins` approach and the user is responsible for specifying the order in which things are processed.

keg/generator.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    # Sanity check image source location
+    if not os.path.exists(os.path.join(recipes_root, 'images', image_source)):
```
rjschwei, line 41: I we make the assumption that the tree lives in git then this implies that we expect the user to have pulled the git repo and has it on their system. This is fair, question I have is whether we should build in a convenience that allows the code to pull the git repo?

- Pro for pulling the git repo is convenience for the user
- Con is that we increase our dependency structure
- Con we'd favor a specific VCS/SCM system (things other than git still exist)

keg/image_definition.py
```diff
@@ -0,0 +1,93 @@
...
+        self._data_root = os.path.join(recipes_root, 'data')
+
+    def populate(self):
+        """Parse recipes data and construct wanted image definition"""
+        utc_now = datetime.now(timezone.utc)
+        utc_now_str = utc_now.strftime("%Y-%m-%d %H:%M:%S")
+        self._data = {
+            'generator': 'keg {}'.format(version.__version__),
+            'timestamp': '{}'.format(utc_now_str),
+            'image source path': '{}'.format(self._image_name)
```
rjschwei, line 41: Singular problem, we need to account for multiple source trees for generating the definition. Also it would be good to have an introductory comment `This kiwi configuration generated by keg......` rather than just `keg`

keg/keg.py
```diff
@@ -15,15 +15,70 @@
...
 def main():
-    log.info(
-        'This is keg version {0}'.format(__version__)
+    argparser = argparse.ArgumentParser(
+        description='Create KIWI image description from keg recipe'
+    )
+    argparser.add_argument(
+        '-r', '--recipes-root',
+        dest='recipes_root',
+        help='Root directory of recipes (required)'
```
rjschwei, line 38: Needs to support multiple, i.e. we either have multiple `--recipes-root` or one with a comma separated list. The help need to be explict about his.

keg/yaml_utils.py
```diff
@@ -0,0 +1,67 @@
...
+# along with keg. If not, see <http://www.gnu.org/licenses/>
+#
+
+from glob import glob
+from pathlib import Path
+import os
+import yaml
+
+
+def _get_versioned_source_files(src_path, include_paths):
```
rjschwei, line 25: Should be recursive or use os.walk() to eliminate explicit tracking of where we have been

rjschwei (changes requested):

rjschwei:
W.r.t. the argument parsing I am partial to argparse, but have no strong preference. While implementation wise the argparse syntax is somewhat archaic I also do not find the docopt style intuitive and always have to use the life-tester on the docopt side to sort out what the syntax should be. Either way it's not a blocker for me.

There are some docstrings missing and some  structural problems, i.e. the singular handling of the description tree. One requirement is that we handle multiple source trees and merge things.

keg/generator.py
jgleissner: `recipes_root` specifies where the data resides and `image_source` specifies which image description keg should generate, which refers to a sub directory within the data tree. Those are two distinct input parameters.

Re multiple `recipes_root` directories, I thought we might be able to use git subtree to pull the private bits from a separate repository into the data tree, although I have to admit that I haven't explored git subtrees much yet. I guess which approach to take would also depend to some degree on whether we want to make keg repository-aware or not.

keg/generator.py
jgleissner: If you use keg manually you may forget to update your tree and work on outdated data, so I guess it wouldn't be just added convenience. Having this as at least an optional feature could be nice.

keg/image_definition.py
jgleissner: The template produces a comment like that:
```
<!-- Image description generated by {{ data['generator'] }} on {{ data['timestamp'] }} -->
```

Re `image source path`, this is really just the name of the image definition and would read something like `sles/on-demand/15-sp2`. It's also not currently used in the templates.

keg/yaml_utils.py
jgleissner: The tracking of where we have been is to avoid importing the same file twice at a different (and unwanted) point. For instance, SLE15-SP2 image definitions uses include paths `sle15-sp2` and `sle15-sp1`. Besides others, the following directories are scanned for yaml files:
```
data/csp/ec2
data/csp/ec2/sle15
data/csp/ec2/sle15/sp1
data/csp/ec2/sle15
data/csp/ec2/sle15/sp2
```
The order how the files are added to the list is important as the most specific ones need to be merged last. However, since `data/csp/ec2/sle15` is twice in the list, so anything included from `data/csp/ec2/sle15/sp1` could potentially be overwritten. Hence the explicit check.

I've tried to make it somewhat more elegant and also looked into using os.walk() but couldn't make it work.

keg/generator.py
rjschwei: I would rather not tie ourselves to a specific RCS, while git is certainly the dominant solution for open source development there are plenty of other options out there and being used.

keg/generator.py
rjschwei: Yes, I think RCS awareness will be a nice feature. We have to design some plugin system such that we do not force the use of a specific RCS system. We would write a git implementation as that is what we use.

keg/image_definition.py
rjschwei: I think more information is valuable thus including `sles/on-demand/15-sp2` in the comment for example would be good.

jgleissner:
I've converted cli parsing from argparse to docopt. It took me a few attempts to get the format right, so I see Robert's argument about it not being very intuitive, but it is more compact and readable so I went with it.

keg/generator.py
```diff
@@ -0,0 +1,86 @@
+# Copyright (c) 2020 SUSE Software Solutions Germany GmbH. All rights reserved.
```
rjschwei, line 1: It's 2021 now ;)

keg/image_definition.py
```diff
@@ -0,0 +1,94 @@
+# Copyright (c) 2020 SUSE Software Solutions Germany GmbH. All rights reserved.
```
rjschwei, line 1: s/2020/2021/

keg/template_functions.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) 2020 SUSE Software Solutions Germany GmbH. All rights reserved.
```
rjschwei, line 1: s/2020/2021/

keg/utils.py
```diff
@@ -0,0 +1,87 @@
+# Copyright (c) 2020 SUSE Software Solutions Germany GmbH. All rights reserved.
```
rjschwei, line 1: s/2020/2021/

rjschwei (approved):
Time to get the first round in and then refine.

jgleissner:
Merging this even though CI checks fail. I was planning to add unit tests beforehand but I want to add root overlay support first and I think this should be a separate PR to not make this one even more convoluted.

PR #6, opened 2021-01-27
Move to github actions

schaefi:
Due to the strange travis policies we are moving the
tox targets to github actions


rjschwei:
LGTM

Issue #10, opened 2021-02-24
Refactor generator.py


schaefi:
The code has room for improvement and does all in one place. In my opinion this needs to be refactored to come in smaller portions such that also tests can be written for it

PR #14, opened 2021-02-25
Test and update image definition

### LINKED ISSUE #11: write tests for ImageDefinition
* write tests for ImageDefinition class
* create test data tree for the tests to run
* update code to meet python quality standards

schaefi:
This patch is many fold but reviewable on a per commit basis

**Added test data infrastructur**
    
For testing components of keg we need a test data infrastructure. The proposed data is based on composing a Leap 15.2 JeOS image. The tests when you read them shows how I understand the purpose of the composing code and its modes

**Add tests for KegImageDefinition class**
    
The KegImageDefinition class takes a keg image definition and composes/includes recursively all data that belongs to this image definition. The component data is taken  from the provided data tree. This commit tests the  traversal of data and checks on the expected output.  This Fixes #11

**Temporary disable 100% code coverage in tests**

Due to the unfortunate merge of code that was not covered by tests the github actions integration reported failures in main which prevents and further pull request to be checked. Until we are writing tests to fully cover the existing code this commit disables the test coverage check such that we can use the pull request checks. NOTE: this will be reverted as soon as we are done with the issue list

**Fixed main docopt setup and use of logging**
    
The logging facility was used in a wrong way by passing the log object around. This commit fixes this part and also puts the docopt string to the python top scope were  it belongs to. The result of this code is now also covered  by tests which makes sure to handle errors the way we  want it

**Flake fixes for keg generator**
    
prior refactor make sure at least the python linter is happy with it

### DIFF
--- a/keg/generator.py
+++ b/keg/generator.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with keg. If not, see <http://www.gnu.org/licenses/>
 #
+import logging
 from jinja2 import Environment, FileSystemLoader
 import os
 
@@ -26,13 +27,13 @@ from keg import (
 )
 from keg.exceptions import KegError, KegKiwiValidationError
 
+log = logging.getLogger('keg')
 
-def create_image_description(image_source,
-                             recipes_root,
-                             data_roots,
-                             dest_dir,
-                             log,
-                             force=False):
+
+def create_image_description(
+    image_source, recipes_root, data_roots,
+    dest_dir, force=False
+):
     """
     Create a KIWI image description in dest_dir from image_source.
 
--- a/keg/generator.py
+++ b/keg/generator.py
@@ -75,7 +75,7 @@ def create_image_description(
         else:
             raise
 
-    script_roots = [ os.path.join(recipes_root, 'data') ] + data_roots
+    script_roots = [os.path.join(recipes_root, 'data')] + data_roots
     script_lib = utils.load_scripts('scripts', script_roots, img['include-paths'])
     config_templ = env.get_template('{}.config.sh.templ'.format(img['schema']))
     config_sh = config_templ.render(data=img.get_data(), scripts=script_lib)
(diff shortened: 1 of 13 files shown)


schaefi:
@jesusbv please ignore the github actions results. Those runs on a per commit base and only the collection of all commits makes the test pass and everything else. Thus when this one is merged we are back on green. Until then please use the

* tox -e check
* tox -e unit_py3_6

calls from the branch of this pull request to check if everything is ok

Thanks

schaefi:
@jesusbv please review on a per commit base. Thanks

keg/image_definition.py
```diff
@@ -15,25 +15,39 @@
...
-    """Class for constructing a keg image definition from recipes"""
-
-    def __init__(self, image_name, recipes_root, data_roots):
-        """Init ImageDefintion with image_name and recipes root path"""
+class KegImageDefinition:
+    """
+    Class for constructing a keg image definition from recipes
+    """
+    def __init__(
+        self, image_name: str, recipes_root: str, data_roots: List = []
```
jesusbv, line 36: do we know the type of the content for `data_roots` list? Probably `List[str]`

keg/generator.py
```diff
@@ -26,13 +27,13 @@
...
 
-def create_image_description(image_source,
-                             recipes_root,
-                             data_roots,
-                             dest_dir,
-                             log,
-                             force=False):
+
+def create_image_description(
+    image_source, recipes_root, data_roots,
```
jesusbv, line 34: no type hints ?

keg/keg.py
```diff
@@ -49,16 +36,28 @@ def main():
         Destination directory for generated description, default cwd
 
     -f, --force
-        Force mode (ignore errors, overwrite files)
+       Force mode (ignore errors, overwrite files)
```
jesusbv, line 39: is it clear that `ignore errors` will log the warnings ?

keg/image_definition.py
schaefi: yes I think this can be done

schaefi: done

keg/generator.py
schaefi: I touched generator only to make it flake clean such that we don't get complains. Let us process all the refactoring for generator.py when we work on that issue #10 ok ?

jesusbv (approved):

PR #18, opened 2021-02-26
Add information for contributing

### LINKED ISSUE #13: Extend README with developer information
The project is setup to use tox for creating dev envs. There is no information for developers how to start with the project. This needs to be added.

jesusbv:
This Fixes #13

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -9,3 +9,54 @@ KEG - Image Compositon Tool
 keg is a command line tool that creates a
 [kiwi](https://github.com/OSInside/kiwi) image description based on
 description snippets in a given GIT repository.
+
+
+## Contents
+
+  * [Contributing](#contributing)
+
+## Contributing
+
+keg is written in Python, it uses [tox](https://tox.readthedocs.io/en/latest/) to setup a development environment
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -1,11 +1,59 @@
...
+keg is written in Python, it uses [tox](https://tox.readthedocs.io/en/latest/) to setup a development environment
+for the desired Python version. Make sure the Python development headers
+are installed (e.g. `python36-devel`)
+
+The following procedure describes how to create such an environment:
+
+1.  Let tox create the virtual environment(s):
+
+    ```
+    $ tox
```
schaefi, line 26: could be ```tox -e devel```  just tox does everything also the docs and the checks and all of it. I think it would be nice if we mention the tox targets that exists. Thoughts

schaefi (approved):
nice, I like it. Just an idea to explain the tox targets we have so people know about "check", "unit_py3_X with X = 6 or 8" and the "doc" target

README.rst
jesusbv: Done

schaefi (approved):
perfect :+1:

PR #20, opened 2021-02-26
Refactor generator

### LINKED ISSUE #10: Refactor generator.py

schaefi:
This change is many fold please review on a commit base

**Refactor generator**
    
Turn the generator into a class. Apply the single responsibilty OO pattern. Add tests for the code and give the single parts of the processing a name. Update the keg commandline code to make use of the classes. Keg should also be usuable as an API,
thus I think it's important come up with a good class design. This Fixes #10

**Added anymarkup to virtual environment**
    
keg will also support YAML output. For this the anymarkup  module is needed.

**Added support for YAML markup in KiwiDescription**


schaefi:
@jesusbv if you review the Generator I suggest you look at the complete code in generator.py. I moved this into a class and also changed the design such that this code can be used easier in an API and is from my pov easier to read and to understand. That causes the diff to be hard to read. Thus just take a look at the complete file. Thanks

keg/generator.py
```diff
@@ -19,69 +19,92 @@
...
-        else:
-            raise KegError(errmsg)
+    def __init__(
+        self, image_definition: KegImageDefinition, dest_dir: str
+    ):
+        self.image_definition = image_definition
+        self.description_schemas = os.path.join(
+            image_definition.recipes_root, 'schemas'
+        )
+        self.dest_dir = dest_dir
```
rjschwei, line 46: IMHO that dest_dir is in fact a directory should be checked 

if no os.path.isdir(dest_dir):
  COMPLAIN

schaefi: yes fully agree and done

keg/generator.py
```diff
@@ -19,69 +19,98 @@
...
+        kiwi_document = kiwi_template.render(
+            data=self.image_definition.data
+        )
+        with open(outfile, 'w') as kiwi_config:
+            kiwi_config.write(kiwi_document)
+        kiwi = KiwiDescription(outfile)
+        if markup == 'xml':
+            kiwi.create_XML_description(outfile)
+        if markup == 'yaml':
+            kiwi.create_YAML_description(outfile)
```
jesusbv, line 84: I think we  should add an `else:` option to log | print a message informing that that markup is not supported.

schaefi: yes I think it should raise an error in this case

schaefi: done

jesusbv (approved):
LGTM

PR #21, opened 2021-02-28
Setup for pypi

schaefi:
**Rename python namespace from keg to kiwi_keg**
    
The project and namespace keg is already registered on the  python package index. Thus we need to rename the python namespace as it cannot co-exist with an already registered one


schaefi:
@jesusbv something we should have checked before but now it was too late. I moved the code and also registered the project on pypi such that this namespace can not be used by others. The github project name we will not change as this makes it even more complicated.

This one should go in first in the series of pull requests and before we continue changing more code as the namespace is now
```from kiwi_keg....``` and no longer ```from keg....``` as this one already exists.

Thanks

jesusbv (approved):

PR #23, opened 2021-03-01
Add template functions tests

### LINKED ISSUE #12: write tests for template_functions
Given template functions should be called through unit tests

jesusbv:
- Test template functions
- Add type hints
- Rename parse_yaml_tree method to get_yaml_tree

This Fixes #12


kiwi_keg/utils.py
```diff
@@ -53,7 +56,8 @@ def _get_source_files(root, src_path, ext, include_paths=None):
     return src_files
 
 
-def rmerge(src, dest):
+def rmerge(src: Dict[str, str], dest: Dict[str, str]) -> Dict[str, str]:
+    """Merge two dictionaries."""
```
schaefi, line 60: can we stick to the comment style which causes no problems to create api docs

```python
"""
Merge two dictionaries
"""
```

And the merge of two dicts is a one liner in python (dict.update) Thus if the comment is correct the extra function would be not needed and otherwise (which is the case) the comment should provide a little bit more context :)
What do you think ?

schaefi (commented):
Like it, just nitpick

schaefi:
Regarding the plain review. Any reason why we should not move this code into a class

```
KegUtils
```

I mean this feels natural and gives the methods a name space. Actually they only work in the context of Keg

Thoughts ?

jesusbv:
> Regarding the plain review. Any reason why we should not move this code into a class
>

No reason other than keep it separate.
I think it should be a class, but I thought it should be done in a different PR
so we reach 100% coverage with this one, and another one for refactor.

schaefi:
agreed

schaefi (approved):

PR #24, opened 2021-03-01
Add type hint checker

### LINKED ISSUE #16: Add mypy type hinting check to tox
keg is being used as an API most probably. Thus after refactor the tox target which runs the tests should also include a successful call for mypy to check the type annotations to be correct

jesusbv:
This Fixes #16


schaefi (approved):
great :+1:

PR #26, opened 2021-03-02
Improve overall error checking

schaefi:
Make sure obvios errors like recipe path does not exist
or schema template was not found are catched by keg.
Also make sure validation errors from KIWI are part of
the log output


jesusbv (approved):
LGTM

PR #27, opened 2021-03-02
Refactor utils

jesusbv:
Turn the code into a class.
Remove code duplication in some methods
Update a method name


schaefi (approved):
Looks great :+1:

PR #28, opened 2021-03-02
Update image definition populate methods

jesusbv:
- Add update profiles method
- Add update contents method

### DIFF
--- a/kiwi_keg/image_definition.py
+++ b/kiwi_keg/image_definition.py
@@ -57,6 +57,10 @@ class KegImageDefinition:
     def data(self) -> Dict:
         return self._data
 
+    @data.setter
+    def data(self, new_data: Dict):
+        self._data = new_data
+
     @property
     def recipes_root(self) -> str:
         return self._recipes_root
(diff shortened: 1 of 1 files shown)


kiwi_keg/image_definition.py
```diff
@@ -78,36 +78,41 @@ def populate(self) -> None:
...
-                        self._data_roots,
-                        include_paths
-                    ),
-                    contents
-                )
-            self._data['contents'].update(contents)
+            self.update_contents(self._data.get('include-paths'))
+
+    def update_profiles(self, include_paths):
+        for profile_name, profile_data in self._data['profiles'].items():
```
schaefi, line 90: This will fail with a stack trace if no 'profiles' key exists. I'd say it should only run when ```if 'profiles' in self.data```

Also I suggest to use the property (data) and not a direct read on _data

kiwi_keg/image_definition.py
```diff
@@ -78,36 +78,41 @@ def populate(self) -> None:
...
+                            ),
+                            profile
+                        )
+                else:
+                    utils.rmerge({item: value}, profile_data)
+            self._data['profiles'][profile_name].update(profile)
+
+    def update_contents(self, include_paths):
+        contents: Dict = {}
+        for inc in self._data['contents'].get('include'):
```
schaefi, line 109: Same here, will fail with a KeyError if no 'contents'

kiwi_keg/image_definition.py
```diff
@@ -78,36 +78,41 @@ def populate(self) -> None:
...
+            utils.rmerge(
+                utils.get_yaml_tree(
+                    inc,
+                    self._data_roots,
```
schaefi, line 113: property access ```self.data_roots```

kiwi_keg/image_definition.py
```diff
@@ -78,36 +78,41 @@ def populate(self) -> None:
...
+                    include_paths
+                ),
+                contents
+            )
+        self._data['contents'].update(contents)
```
schaefi, line 118: can also fail with a KeyError on 'contents'

schaefi (changes requested):
See my comments

kiwi_keg/image_definition.py
jesusbv: > This will fail with a stack trace if no 'profiles' key exists. I'd say it should only run when `if 'profiles' in self.data`
>
Done.  

> Also I suggest to use the property (data) and not a direct read on _data
>
Good catch, using the properties now, added two more and a setter as well.

schaefi (approved):
like it :+1:

schaefi:
@jesusbv looks good, it just conflicts now with the other PR that was merged. Do you want me to solve the conflict ?

PR #33, opened 2021-03-04
Preserve XML comments

### LINKED ISSUE #19: Preserve XML comment blocks
We use the kiwi API to create and validate the keg created description. On that processing XML comment blocks will not be translated. Thus we need to preserve the comment based header and add it back because this information is not a comment for the open buildservice

schaefi:
The KIWI validation does not preserve comments after validation
as they have no meaning for the processing of the image description.
However for OBS comments are treated as project configurations
and therefore the toplevel comments if present must be added back.
This Fixes #19

### DIFF
--- a/kiwi_keg/generator.py
+++ b/kiwi_keg/generator.py
@@ -46,6 +46,12 @@ class KegGenerator:
                     target=repr(dest_dir)
                 )
             )
+        self.kiwi_description = os.path.join(
+            dest_dir, 'config.kiwi'
+        )
+        self.kiwi_config_script = os.path.join(
+            dest_dir, 'config.sh'
+        )
         self.image_definition = image_definition
         self.description_schemas = os.path.join(
             image_definition.recipes_root, 'schemas'
@@ -69,17 +75,14 @@ class KegGenerator:
             )
         )
 
-    def create_kiwi_description(
-        self, markup: str = 'xml', override: bool = False
-    ):
+    def create_kiwi_description(self, override: bool = False) -> None:
         """
         Creates KIWI config.xml from a KegImageDefinition.
 
         :param bool override:
             Override destination contents, default is: False
         """
-        outfile = os.path.join(self.dest_dir, 'config.kiwi')
-        self._validate_outfile(outfile, override)
+        self._check_file(self.kiwi_description, override)
 
         kiwi_template = self._read_template(
             '{}.kiwi.templ'.format(self.image_schema)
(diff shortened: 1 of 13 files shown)


kiwi_keg/generator.py
```diff
@@ -142,3 +155,19 @@ def _read_template(self, template_name):
...
+                if comment.startswith(('<!--', '-->', '<?')):
+                    comments.append(comment)
+        if comments:
+            with open(xml_file, 'r') as xml:
+                xml_data = xml.read()
+
+            with open(xml_file, 'w') as xml:
+                xml.write(os.linesep.join(comments))
+                xml.write(os.linesep)
+                xml.write(xml_data)
```
rjschwei, line 173: Hmm, this is somewhat unrelated to this work but still something we might want to consider. At present our files have section comments such as  looks like we loose that data. Given that what's in the build service will still be read by humans it would really be great to figure out a way to maintain those guiding comments as well.

kiwi_keg/generator.py
```diff
@@ -92,6 +92,19 @@ def create_kiwi_description(
...
+            # influence this and add back the header comments.
+            #
+            # This is only done for the XML markup since I think
+            # OBS is expecting the comments in XML comment syntax
+            # and does not support anything else. If this turns
+            # out to be an issue in the future and people start
+            # to use YAML markup for KIWI images in OBS we need
+            # add a translation of the XML comments to YAML
+            # after create_YAML_description
+            self._preserve_XML_comments(kiwi_document, outfile)
```
jesusbv, line 107: I wonder if `KegGenerator` should preserve the comments or it is a `KiwiDescription responsability`
Maybe after this [line](https://github.com/SUSE-Enceladus/keg/blob/main/kiwi_keg/kiwi_description.py#L64)
so it could be 
```python
def create_XML_description(self, output_file: str):
        self._create_description(output_file, 'xml')
        self._preserve_XML_comments(params)
```

schaefi: you are right it's a KiwiDescription responsibility. I moved the code. Thanks much

kiwi_keg/generator.py
schaefi: I'm looking into it

schaefi: Put back inline comments has many many problems. I tried almost all day to come up with a good solution and failed.

* You need an anchor in the document to put back an inline comment at the right place. line numbers does not work
   because they are different before and after the transformation

* The current implementation is super silly and just uses the previous line as anchor information but this might be wrong as the same information can exist at other places in the document. So worst case the comment is placed at the wrong place

* The XSLT transformation updates the schema. This means attributes and sections changes. If the anchor was an information that got changed by the transformation you will not find this position at all. So the comment gets lost

Overall this is the problem of diff/patch and there are very sophisticated algorithms to find the right place in the document but all that would be way too much overhead. Using patch is also not an option because it will fail if the XSLT transfornation has shifted the document to another schema.

I'm sorry my vote is to care for the header comments which can be placed safely and give up on inline comments

And another option would be to add a keg option that skips the schema update and allow users to do the move when it is required but not force this operation

Thoughts ?

schaefi: I will handle this via extra options

jesusbv: I think the root of the issue is XSLT transforming the document, so fix should be there but that does not seem to be possible
I agree, better not to offer something that we know is going to be potentially wrong

schaefi:
ok I pushed a commit which allows to use keg a bit differently as before. This was done to handle the comment situation more explicit and also show information about it to the user. With the current changes, Keg by default only presents the XML version which is the result of the jinja template and validates this data against the kiwi API

Using the new options ```--format-xml or --format-yaml``` allows to format/update the written Keg description with the installed KIWI schema and validate this result. Using this options impacts the transformation of eventually existing comment blocks in the Keg written description as we know. Therefore a warning message is displayed which says for ```format-xml``` that only the toplevel comments will be preserved and in case of ```--format-yaml``` comments are currently not handled at all

This is from my perspective a good change to allow most flexibility and to improve on comment handling when the time comes

Thoughts ?

Thanks

jesusbv (approved):
LGTM

PR #34, opened 2021-03-07
Add support for overlay files

### LINKED ISSUE #30: Add support for overlay files
Currently keg does not handle overlay files. We need a design and agreement on how to manage this information in a keg image definition tree

jesusbv:
This Fixes #30


schaefi:
Hi, I did not quite understand the processing of files and the copy. So I decided to just give it a real world test

I've checked out keg-recipes and added some overlay files. Looks like this

```
find overlays/
overlays/
overlays/base
overlays/base/etc
overlays/base/etc/motd
overlays/products
overlays/products/leap
overlays/products/leap/15.2
overlays/products/leap/15.2/usr
overlays/products/leap/15.2/usr/lib
overlays/products/leap/15.2/usr/lib/systemd
overlays/products/leap/15.2/usr/lib/systemd/system
overlays/products/leap/15.2/usr/lib/systemd/system/foo.service
```

Then I updated images/leap/jeos/15.2/image.yaml as follows

```yaml
include-paths:
  - base/jeos/leap15
overlay-include-paths:
  - base
  - products/leap/15.2
image:
  name: Leap15.2-JeOS
  specification: "Leap 15.2 guest image"
  version: 1.15.2
repositories:
  -
    path: obs://openSUSE:Leap:15.2/standard
```

then I called keg

```
keg --recipes-root ~ms/Project/keg-recipes/ --dest-dir leap_description leap/jeos/15.2 --force
```

and it failed with

```
[ ERROR   ]: 16:48:39 | Unexpected error:
Traceback (most recent call last):
  File "/home/ms/Project/keg/.tox/3.6/bin/keg", line 33, in <module>
    sys.exit(load_entry_point('kiwi-keg', 'console_scripts', 'keg')())
  File "/home/ms/Project/keg/kiwi_keg/keg.py", line 82, in main
    image_generator.create_overlays()
  File "/home/ms/Project/keg/kiwi_keg/generator.py", line 144, in create_overlays
    shutil.copy(name, dest_file)
  File "/home/ms/Project/keg/.tox/3.6/lib64/python3.6/shutil.py", line 245, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/home/ms/Project/keg/.tox/3.6/lib64/python3.6/shutil.py", line 121, in copyfile
    with open(dst, 'wb') as fdst:
FileNotFoundError: [Errno 2] No such file or directory: 'leap_description/etc/motd'
```

So did I something wrong ?

Thanks

schaefi:
Thanks Jesus I tested again and this time it produced the files as expected. I realized the files are created directly in the given dest-dir. So it looked like this:

```
keg --recipes-root ~ms/Project/keg-recipes/ --dest-dir leap_description leap/jeos/15.2 --force
[ INFO    ]: 08:33:38 | Using KIWI schema: vm
[ INFO    ]: 08:33:38 | Support for multiple markup descriptions available
ms@asterix:/home/ms/Project/keg
(include-overlays)> cd leap_description/
ms@asterix:/home/ms/Project/keg/leap_description
(include-overlays)> ll
total 16K
-rw------- 1 ms users 2.1K Mar  9 08:33 config.kiwi
-rw-r--r-- 1 ms users 2.9K Mar  9 08:33 config.sh
drwxr-xr-x 2 ms users 4.0K Mar  9 08:33 etc
drwxr-xr-x 3 ms users 4.0K Mar  9 08:33 usr
```

schaefi:
The overlay files needs to be organized in a root/ directory below --dest-dir. Can you change your code in that direction ?

schaefi:
Also please add an option to keg which allows to create a tarball named ```root.tar.gz``` instead of the files as root/ directory

jgleissner:
Hi guys, thanks for working on this while I was out. I had basically finished the keg-recipes part of the overlay support just before my accident. Too bad I thought "ah well I'll open that PR on Monday". ;) Anyway, the approach I had in mind is similar but also a bit different in the way that I thought it would make sense to group files logically into features and then include those features in the data description. I guess it's easiest to open a PR in keg-recipes with what I have and we can take it from there.

schaefi:
Hey Joe, great to hear from you :smiley: 

> I guess it's easiest to open a PR in keg-recipes with what I have and we can take it from there.

yes I think this works best. I'd like to finish this one up and get it merged and then we can talk about the data as you provide it in another PR on keg-recipes. If handling the data needs a change on the keg code we can come up with a PR here too.

Hope you are well and we are happy to have you back :)

schaefi:
@jesusbv I tested the tarball and there is a small issue.

It lists as

```
drwxr-xr-x ms/users          0 2021-03-09 13:31 root/
drwxr-xr-x ms/users          0 2021-03-09 13:31 root/etc/
-rw-r--r-- ms/users          4 2021-03-10 10:58 root/etc/motd
drwxr-xr-x ms/users          0 2021-03-09 13:31 root/usr/
drwxr-xr-x ms/users          0 2021-03-09 13:31 root/usr/lib/
drwxr-xr-x ms/users          0 2021-03-09 13:31 root/usr/lib/systemd/
drwxr-xr-x ms/users          0 2021-03-09 13:31 root/usr/lib/systemd/system/
-rw-r--r-- ms/users          4 2021-03-10 10:58 root/usr/lib/systemd/system/foo.service
```

but it has to be

```
drwxr-xr-x ms/users          0 2021-03-09 13:31 etc/
-rw-r--r-- ms/users          4 2021-03-10 10:58 etc/motd
drwxr-xr-x ms/users          0 2021-03-09 13:31 usr/
drwxr-xr-x ms/users          0 2021-03-09 13:31 usr/lib/
drwxr-xr-x ms/users          0 2021-03-09 13:31 usr/lib/systemd/
drwxr-xr-x ms/users          0 2021-03-09 13:31 usr/lib/systemd/system/
-rw-r--r-- ms/users          4 2021-03-10 10:58 usr/lib/systemd/system/foo.service

```

Can you please fix that ? Thanks

jgleissner:
I don't really like having a separate include directive in the image configuration. We already include modules based on the functionality they provide (e.g. `csp/ec2` provides package and config data to support EC2), no reason why the files overlay should require its own configuration. I would rather add the overlay files to the data modules, and keep the image meta configuration as straight-forward as possible. See https://github.com/SUSE-Enceladus/keg-recipes/pull/5 for my proposal for the data structure.

schaefi:
> We already include modules based on the functionality they provide (e.g. csp/ec2

I'm not sure if the logic for including modules is always the same as with including overlay files. It can happen relatively easily that the include path for a module must be changed to differentiate a certain property but stays the same for the overlay files.
I'm ok if we try to use the same information from ```include-paths``` also for collecting the overlay files but I'm not sure if we will be able to sustain it

Regarding https://github.com/SUSE-Enceladus/keg-recipes/pull/5 I only have one concern. You put all overlay files at ```data/files``` and you reference them with ```files:``` in the yaml. I think it should be super clear that we are talking about overlay files because there is already a files section and that is being used for the scripts.

So please let us rename that to ```data/overlayfiles``` and also ```overlayfiles:``` in the yaml.

If we go that way I'm fine and we can try to use ```include-paths``` for modules and overlay files, even though I think these are two different things

jgleissner:
> > We already include modules based on the functionality they provide (e.g. csp/ec2
> 
> I'm not sure if the logic for including modules is always the same as with including overlay files. It can happen relatively easily that the include path for a module must be changed to differentiate a certain property but stays the same for the overlay files.
> I'm ok if we try to use the same information from `include-paths` also for collecting the overlay files but I'm not sure if we will be able to sustain it

I think we need to be more clear about the terminology.  There is the `include` directive in the image definition that is there to include a certain functionality in the image (JeOS, SLES, CSP specifics, etcs). And there is the `include-paths` directive, which is used to define path trees that get appended to all directories when scanning the data modules, which is there to allow overriding configuration inherited from more upper levels by more specific configurations bits and also to enable cross SP inheritance.

The overlay functionality as implement in this PR add another set of directories added to the image definition, but without the inheritance mechanic. It means that you have to configure the overlay in every image definition and potentially have to edit several ones in case things change, which is one thing we wanted to avoid in the first place.

So, I don't think we need to configure overlay files in the image definition. The content in the overlay tarballs differ very little in terms of functional groups between products, SPs, and even SLES versions, which leads me to believe that handling the overlay files the same as the rest of the image configuration should work, and should provide for least duplication. I'll add the overlay configurations next to the recipes, which should prove (or disprove) this.

> Regarding [SUSE-Enceladus/keg-recipes#5](https://github.com/SUSE-Enceladus/keg-recipes/pull/5) I only have one concern. You put all overlay files at `data/files` and you reference them with `files:` in the yaml. I think it should be super clear that we are talking about overlay files because there is already a files section and that is being used for the scripts.
> 
> So please let us rename that to `data/overlayfiles` and also `overlayfiles:` in the yaml.

Yes, it should be namespaced under `overlay` in the yaml config analog to config. Sorry, fixed that. Not sure about renaming the directory though. I don't think it adds anything useful but I don't feel strongly about it.

schaefi:
>The overlay functionality as implement in this PR add another set of directories added to the image definition, but without the >inheritance mechanic.

I agree to that, the inheritance model should apply to the overlay files too

> So, I don't think we need to configure overlay files in the image definition

Actually I don't care much from which place the info is picked up. There is also a singlebuild schema which I though was used with just one yaml definition and a flat hierarchy. I'm fine with the place you chose to add the selection of overlay files.

>Yes, it should be namespaced under overlay in the yaml config analog to config. Sorry, fixed that. Not sure about renaming the >directory though. I don't think it adds anything useful but I don't feel strongly about it.

I have stronger feelings on that part though. Please remember we jumped in and had to understand the data structure as beginners. This is an advantage because from there things are not as obvious as they are for you. I'd like to have this
more clear. I picked the following from the open PR on keg-recipes:

```yaml
overlay:
  files:
    EC2:
       - cloud-cfg-ec2
```

why so nested ? why not simply

```yaml
overlayfiles:
  EC2:
    - cloud-cfg-ec2
```

and then

```
data/files/boto-gce-hack/etc/init.d/boot.local
```

why so inconsistent ? why not

```
data/overlayfiles/boto-gce-hack/etc/init.d/boot.local
```

We should do this right in this phase of the project, we will never have the chance to change this when the first version gets out. I have identified several of these sort of issues and the code and the data structures are much easier to read and to understand if the key names are clear and the nesting is as flat as possible. You said this does not add anything useful but I think data consistency and flat structures are a great value to allow for better understanding and maybe also change

Just my 2cents on data structures

jgleissner:
> > Yes, it should be namespaced under overlay in the yaml config analog to config. Sorry, fixed that. Not sure about renaming the >directory though. I don't think it adds anything useful but I don't feel strongly about it.
> 
> I have stronger feelings on that part though. Please remember we jumped in and had to understand the data structure as beginners. This is an advantage because from there things are not as obvious as they are for you. I'd like to have this
> more clear. I picked the following from the open PR on keg-recipes:
> 
> ```yaml
> overlay:
>   files:
>     EC2:
>        - cloud-cfg-ec2
> ```
> 
> why so nested ? why not simply
> 
> ```yaml
> overlayfiles:
>   EC2:
>     - cloud-cfg-ec2
> ```

My idea was to have it consistent with the other sections (packages, config) which all structured as follows:

```
section:
  param-type:
    namespace:
```

I realize it's not very flat in nature, but kinda necessary for the way I wanted to handle the data. As for the overlay, not sure whether other parameter types than `files` would be necessary or useful. Perhaps not.

> and then
> 
> ```
> data/files/boto-gce-hack/etc/init.d/boot.local
> ```
> 
> why so inconsistent ? why not
> 
> ```
> data/overlayfiles/boto-gce-hack/etc/init.d/boot.local
> ```

I don't see the inconsistency. It's files that get injected into the image. I'm happy keeping them in a directory names just `files` but if people like `overlayfiles` better, fine with me.

> We should do this right in this phase of the project, we will never have the chance to change this when the first version gets out. I have identified several of these sort of issues and the code and the data structures are much easier to read and to understand if the key names are clear and the nesting is as flat as possible. You said this does not add anything useful but I think data consistency and flat structures are a great value to allow for better understanding and maybe also change

Absolutely. Please raise any concerns you have. It's all meant to be discussed. :)

rjschwei:
> The overlay files needs to be organized in a root/ directory below --dest-dir. Can you change your code in that direction ?

Hmm is a generic `root` really what we want? For example when we build images with profiles (multibuild) as we do for 15 SP2 collecting everything in $DESTDIR/root would certainly create a mess. I think the destination needs to be configurable or we establish a convention that the destination of the overlay tree as it gets collected is named after the profile. In any event just `root` will not work. `root` should be the default if there are no profiles involved.

rjschwei:
I like the idea of being more explicit, i.e. I would prefer `overlayfiles` over `files`

rjschwei:
With respect to inheritance, it is necessary. We need to be able to override a file that is part of the "generic config" with something that is more specific.

Generally I am in favor of using the data structure in https://github.com/SUSE-Enceladus/keg-recipes/pull/5 as the guide for the implementation plus the discussed clarifications.

schaefi:
> In any event just root will not work. root should be the default if there are no profiles involved

good point, we haven't thought about that. What you are talking about is the difference between root.tar.gz which is handled by OBS and always used when present and the feature in kiwi which allows to say ```<archive name="..."/>```. We have images that uses one or the other feature. From my perspective we have to have a flag in the image definition tree that explicitly specify the use of the archive feature. Connecting this with the simple presence of a profile is in my opinion not so nice because there is no direct relation between profiles and archives. We designed our images that this relation exists but it's only our special image design

schaefi:
> My idea was to have it consistent with the other sections (packages, config) which all structured as follows:

I guessed this yes and you use this information in the templates which is why the structure becomes important at that
place. BUT; overlayfiles are not handled by any template. This information will only be handled by the keg code to create
an overlay file tree or tar archive(s). Because of that there is no reason to be consistent with other structure types and we
can focus on the most simple and clean way to describe this information

> As for the overlay, not sure whether other parameter types than files would be necessary or useful. Perhaps not.

That's my point all you need is

```
overlayfiles:
  EC2:
    - cloud-cfg-ec2
```

and keg would create EC2.tar.gz. If the namespace is set to ```root:``` keg would create root.tar.gz and so on. That way we create all the tarballs. The reference to the ```<archive>``` is already implemented in the ```archive:``` section which you handle in the templates already

from schemas/vm.kiwi.templ
```
{%- for namespace, packages in content['packages'][pkg_type].items() %}
        <!-- Begin include from {{ namespace }} -->
        {%- if namespace == 'archive' %}
        <archive name="{{ packages['name'] }}"/>
```

So from my perspective there is no reason to add a param-type layer for overlayfiles and we should keep things as simple as possible. It's already a pain to work with the .templ files :)

Thanks

jgleissner:
The template files are indeed a bit messy. I've been meaning to make them a bit nicer, either by moving some code into a functions template or preprocess the data in keg and provide it to the template in a form that's a bit easier to handle in jinja. But that wasn't super high on the priority list.

Wrt to the overlay tarballs, I thought we would have one per profile (or just one for multi-build), but I guess we could generate one per overlay files section and simply have a bunch of them. I assume kiwi would be fine with that. That archive snippet in the template doesn't do anything at this point and will need to be adjusted to look up the archive names where we end up putting them, of course.

schaefi:
> Wrt to the overlay tarballs, I thought we would have one per profile

But we should not make that a hard wired requirement. I mean we implement this as opensource project for undefined use cases. Our design "one archive per profile" is just one design. We should not hard wire this concept into the image definition tree.

> That archive snippet in the template doesn't do anything at this point 

This is not true

```yaml
packages:
  image:
    archive:
      name: foo
```

and you get

```
<!-- Begin include from archive -->
    <archive name="foo"/>
<!-- End include from archive -->
```
 
So a description for a simple root.tar.gz tarball could be

```yaml
overlayfiles:
    - abc
```

And a description for overlay files based on an archive could be

```yaml
overlayfiles:
   EC2:
     - abc
     
packages:
  image:
    archive:
      name: EC2
```

Will result in EC2.tar.gz and the respective ```<archive name="EC2"/>``` whereas we need add the tar.gz extension

I have tested this so it's also not needed that we kind a create dependency between profiles and archives

schaefi:
and of course this also works with profiles.

schaefi:
So this conversation is drifting a little bit. I like the setup of the overlay files in SUSE-Enceladus/keg-recipes#5 I just don't like the nested ```overlay -> files``` data structure when it could be just simple. And iirc we all agree that ```data/overlayfiles/...``` is nicer than ```data/files/...```

So I'm aiming for an agreement such that we can go ahead. From a process perspective we should merge SUSE-Enceladus/keg-recipes#5 first such that we have data to work with.

I think I've made my point clear and I also have live tested that it works but there can only be one cook and you are assigned as the tech lead. Thus the last word is on you

Thanks

jgleissner:
True, if you add an `archive` key under `packages:` it will generate that, but it's not terribly useful. If we're adding that information under `overlayfiles:` than that's where we should take it from, that was all I was saying. I also didn't mean to push the one-archive-per-profile idea, I was just describing where I was coming from, but I like the multiple-archives approach actually better, as it seems to be simpler. :)

jesusbv:
have updated the PR with minor changes for CI and renames from yesterday commit plus solved the conflicts.
I have implemented the overlays per profile but it seems that is not the direction we are going, so those changes are not pushed, until there is a clear decision on that.

rjschwei:
> > In any event just root will not work. root should be the default if there are no profiles involved
> 
> good point, we haven't thought about that. What you are talking about is the difference between root.tar.gz which is handled by OBS and always used when present and the feature in kiwi which allows to say `<archive name="..."/>`. We have images that uses one or the other feature. From my perspective we have to have a flag in the image definition tree that explicitly specify the use of the archive feature. Connecting this with the simple presence of a profile is in my opinion not so nice because there is no direct relation between profiles and archives. We designed our images that this relation exists but it's only our special image design

Well the "flag" could simply be `overlayname: ec2` and that would supersede the default `root` implementation.

rjschwei:
And now that I've read everything and taking the "multiple" archives condition into account I think that we definitely want `overlayname` and that would then look like this:

```
overlayfiles:
   EC2:
     overlayname: ec2-1
         - abc
         - foo
     overlayname: ec2-2
         - bar
         - beer
 ```

I think if we find an overlay definition that does not have the default name, i.e. root we should automatically generate the `<archive..../>` entry. But that requires us to create a convention that the entry in `overlayfiles` matches a defined profile. I think that is a reasonable convention.

**Comment by jgleissner** [comment]
I've added (most of) the overlay configurations to keg-recipes. Based on that input, the overlay data should automatically be read by keg into the image description dictionary, and hence be accessible via data['profiles']['profile_name']['overlayfiles']. The idea is to produce a tarball for each key in `overlayfiles` (I guess `keyname.tar.gz` could serve as filename). Each key holds a list of strings which map to directories in `data/overlayfiles`, which hold the content that should go into the tarball.

File: kiwi_keg/generator.py [hunk 1]
```diff
@@ -153,6 +155,81 @@ def create_custom_scripts(self, override: bool = False):
...
+            archive_element = etree.SubElement(image_element, 'archive')
+            archive_element.attrib['name'] = archive_name
+            tree = etree.ElementTree(kiwi_root)
+            tree.write(
+                self.kiwi_description,
+                encoding="utf-8",
+                xml_declaration=True,
+                pretty_print=True
+            )
+
```
schaefi, line 232: If we decide to manage the XML content by keg outside of the template the best way to do it is by using the KIWI API

```python
import sys
import logging

from kiwi.xml_description import XMLDescription
from kiwi.xml_state import XMLState
from kiwi import xml_parse

description = XMLDescription('/home/ms/Project/kiwi-descriptions/suse/x86_64/suse-leap-15.2/config.xml')

xml_data = description.load()

xml_state = XMLState(
    xml_data=xml_data, profiles=['Disk'], build_type='oem'
)

package_section = xml_state.get_packages_sections(
    section_types = ['image']
)[0]

package_section.add_archive(
    xml_parse.archive(name='foo.tar.gz')
)

xml_data.export(
    outfile=sys.stdout, level=0
)
```

be aware this will bring back the comment issue

schaefi: because of the side effects when using the kiwi api (comments and auto update to the schema of the api) I suggest to handle the XML settings via the schema templates as it is done for the other settings as well

jgleissner:
I'm working on updating the overlay data in keg-recipes, as discussed on Friday. To recap, we want to have the option to name archives explicitly but also to produce pain root.tar.gz, since we have builds that use that and we want to be able to reproduce those. Plus it was requested that we you can named archives and root.tar.gz at the same time.

Here's my suggestion for the structure. Assume this is the overlay data keg pieced together based on the config for an Azure image:

```
overlayfiles:
  azure-common:
    include:
      - cloud-cfg-azure
      - cloud-cfg-disable-network-config
      - sysconfig-network-cloud-netconfig
  azure-sle15-sp3:
    include:
      - feature-added-in-sle15-sp3
  azure-extra-stuff:
    name: azure-extra
    include:
      - some-feature
```

Based on that, keg could put anything that does not have an explicit name either into `root.tar.gz` (if shows up under profiles/common, if single-build is used or data is common to all profiles) or `{profile}.tar.gz` (if profile specific). `azure-extra-stuff` would always end up in `azure-extra.tar.gz` regardless of image layout. This allows to separate files into their own archive, for instance if you have some bigger files you want to include in multiple profiles but not all of them. This way keg can add the archive tag to the appropriate profile sections rather than adding the same files to multiple tarballs.

jesusbv:
> I'm working on updating the overlay data in keg-recipes, as discussed on Friday. To recap, we want to have the option to name archives explicitly but also to produce pain root.tar.gz, since we have builds that use that and we want to be able to reproduce those. Plus it was requested that we you can named archives and root.tar.gz at the same time.
> 

yes, but what I understood was agreed on the meeting on Friday was this structure

```yaml
overlayfiles:
     overlayname: root
        - some/path
     overlayname: ec2-1
         - abc
         - foo
     overlayname: ec2-2
         - bar
         - beer
```

jgleissner:
That structure does not work. You can't use the same key (`overlayname` in this case) multiple times. We need to namespace the sections.

jesusbv:
> That structure does not work. You can't use the same key (`overlayname` in this case) multiple times. We need to namespace the sections.

You can, I did it. I'm using it.

```
overlayfiles:
     overlayname: root
         - some/path
     overlayname: ec2-1
         - abc
         - foo
     overlayname: ec2-2
         - bar
         - beer
```
produces
```python
{'overlayfiles: {'overlayname': {'root': ['some/path'], 'ec2-1': ['abc', 'foo'], 'ec2-1': ['bar', 'beer']}
```
in image_definition data.

also, different tarballs are created accordingly

jgleissner:
```yaml
$ cat foo.yaml 
overlayfiles:
     overlayname: root
         - some/path
     overlayname: ec2-1
         - abc
         - foo
     overlayname: ec2-2
         - bar
         - beer
```

```
$ python3 -c "import yaml; print(yaml.safe_load(open('foo.yaml', 'r').read()))"
{'overlayfiles': {'overlayname': 'ec2-2 - bar - beer'}}
```

Not sure what I'm doing differently. Besides, even if it worked like you described, that would make it essentially the same as:

```yaml
overlayfiles:
     overlayname:
         root:
             - some/path
         ec2-1:
             - abc
             - foo
         ec2-2:
             - bar
             - beer
```

Not sure whether I see the need for the `overlayname` level. Plus we will still need to namespace, because `root` will not be unique.

jesusbv:
> $ python3 -c "import yaml; print(yaml.safe_load(open('foo.yaml', 'r').read()))"
> {'overlayfiles': {'overlayname': 'ec2-2 - bar - beer'}}
> 
That is the result applying default loader, one can set a different loader to preserve duplicate keys (or customize it any desired way)
 
> Not sure whether I see the need for the `overlayname` level. Plus we will still need to namespace, because `root` will not be unique.

We agreed on the structure because it works, makes sense and it is explicit. 
So keg users can read the overlay yaml and understand that `root` or `ec2-1` is the overlay name. 
I'm not sure I understand the `root` will not be unique concern.

jgleissner:
> > $ python3 -c "import yaml; print(yaml.safe_load(open('foo.yaml', 'r').read()))"
> > {'overlayfiles': {'overlayname': 'ec2-2 - bar - beer'}}
> 
> That is the result applying default loader, one can set a different loader to preserve duplicate keys (or customize it any desired way)

In preserving duplicate keys you effectively disable the possibility to override configuration inherited from an upper layer. That's not really an option.

> > Not sure whether I see the need for the `overlayname` level. Plus we will still need to namespace, because `root` will not be unique.
> 
> We agreed on the structure because it works, makes sense and it is explicit.
> So keg users can read the overlay yaml and understand that `root` or `ec2-1` is the overlay name.

I find `overlayname` as an additional nesting level rather pointless. I find the `name` parameter I suggested a bit more meaningful. It makes clear that this set of files should have an explicit name, and it does that without adding an unnecessary nesting level.

> I'm not sure I understand the `root` will not be unique concern.

Because keys need to be unique.

rjschwei:
> > > $ python3 -c "import yaml; print(yaml.safe_load(open('foo.yaml', 'r').read()))"
> > > {'overlayfiles': {'overlayname': 'ec2-2 - bar - beer'}}
> > 
> > 
> > That is the result applying default loader, one can set a different loader to preserve duplicate keys (or customize it any desired way)
> 
> In preserving duplicate keys you effectively disable the possibility to override configuration inherited from an upper layer. That's not really an option.

That is only true if we refuse to manage the dict on our own. We can let the loader use "safe load" to preserve duplicates and then decide which pieces may or may not be overwritten on our own.

> 
> > > Not sure whether I see the need for the `overlayname` level. Plus we will still need to namespace, because `root` will not be unique.
> > 
> > 
> > We agreed on the structure because it works, makes sense and it is explicit.
> > So keg users can read the overlay yaml and understand that `root` or `ec2-1` is the overlay name.
> 
> I find `overlayname` as an additional nesting level rather pointless. 

Well but that's what we agreed on. We can have another conversation, but if you find it pointless it would have been great to have that raised in the conversation we had rather than after the fact.

> I find the `name` parameter I suggested a bit more meaningful.

Well I don't necessarily see the benefit of having the same string repeated twice

```
azure-extra-stuff:
    name: azure-extra
    include:
      - some-feature
```

might as well be:

```
azure-extra:
    name: azure-extra
    include:
      - some-feature
```

Why would I have to type that twice?

And then

```
azure-extra:
    name: root
    include:
      - some-feature
```
may or may not be illegal  if

```
foo:
  name: root
  include:
    - beer
```
Also exists or do you propose to just merge the root definitions in that case?

> It makes clear that this set of files should have an explicit name, and it does that without adding an unnecessary nesting level.
> 
> > I'm not sure I understand the `root` will not be unique concern.
> 
> Because keys need to be unique.

Not necessarily, it depends on how we decide to handle this. But obviously we are not on the same page yet so we should have another conversation this week.

jgleissner:
> > In preserving duplicate keys you effectively disable the possibility to override configuration inherited from an upper layer. That's not really an option.
> 
> That is only true if we refuse to manage the dict on our own. We can let the loader use "safe load" to preserve duplicates and then decide which pieces may or may not be overwritten on our own.

Certainly possible. The current implementation uses the fact the a key will overwrite an existing key with the same name, and traverses directories from generic to increasingly specific. If another design is preferable, fine with me.

> > I find `overlayname` as an additional nesting level rather pointless.
> 
> Well but that's what we agreed on. We can have another conversation, but if you find it pointless it would have been great to have that raised in the conversation we had rather than after the fact.

I thought the option to have an explicit way to define the archive name was the point. I didn't think we were talking about that exact syntax.

> > I find the `name` parameter I suggested a bit more meaningful.
> 
> Well I don't necessarily see the benefit of having the same string repeated twice
> 
> ```
> azure-extra-stuff:
>     name: azure-extra
>     include:
>       - some-feature
> ```
> 
> might as well be:
> 
> ```
> azure-extra:
>     name: azure-extra
>     include:
>       - some-feature
> ```
> 
> Why would I have to type that twice?

Somehow you'll have to make clear you want a named archive rather than having the files lumped in `root.tar.gz` or `${profile}.tar.gz`. And as mentioned, to stay with the current design, a namespace is required, so if you insist on having `overlayname: foo` for syntactical reason, we end up with this:

```yaml
overlayfiles:
     azure-common:
         overlayname: root
             - common-feature1
             - common-feeature2
     azure-extra:
         overlayname: azure-extra
             - extra-stuff
```

You'd still have the namespace name and the overlayname. We could have some sort of flag instead of overlayname. It would still require that hackery in https://github.com/SUSE-Enceladus/keg/blob/ec82f39b533165b6167c1cae189e5cc7e0214661/kiwi_keg/utils.py#L186-L189 though because this isn't even proper yaml.

> And then
> 
> ```
> azure-extra:
>     name: root
>     include:
>       - some-feature
> ```
> 
> may or may not be illegal if
> 
> ```
> foo:
>   name: root
>   include:
>     - beer
> ```
> 
> Also exists or do you propose to just merge the root definitions in that case?

I would try to merge those when populating the dict. I'd add a function that iterates over the overlay configurations, figures out which filename to use for each archive, creates `archive` tags for the profiles if necessary, and builds file lists in the process.
 
> > It makes clear that this set of files should have an explicit name, and it does that without adding an unnecessary nesting level.
> > > I'm not sure I understand the `root` will not be unique concern.
> > 
> > 
> > Because keys need to be unique.
> 
> Not necessarily, it depends on how we decide to handle this. But obviously we are not on the same page yet so we should have another conversation this week.

Sure.

schaefi:
I suggest we use the following layout:

```yaml
overlayfiles:
  -
    overlayname: root
      - some/path

  -
    overlayname: ec2-1
      - abc
      - foo

  -
    overlayname: ec2-2
      - bar
      - beer
```

results in

```
{
    'overlayfiles': [
        {
            'overlayname': 'root - some/path'
        },
        {
            'overlayname': 'ec2-1 - abc - foo'
        },
        {
            'overlayname': 'ec2-2 - bar - beer'
        }
    ]
}
```

which I think is what we had discussed and also agreed on, or am I mistaken ?

schaefi:
>I'm not sure I understand the root will not be unique concern.
>>Because keys need to be unique.

The name ```root``` is not a key. It causes keg to create a tarball named ```root.tar.gz```. Then we agreed on a keg convention which is if the overlayname is set to root we don't need an archive statement in the kiwi configuration. I don't see anything wrong with it and also thought we clarified that part

schaefi:
Maybe this can be solved with writing the yaml like this

```yaml
overlayfiles:
  -
    overlayname: 'root'
    elements:
      - 'some/path'

  -
    overlayname: 'ec2-1'
    elements:
      - 'abc'
      - 'foo'

  -
    overlayname: 'ec2-2'
    elements:
      - 'bar'
      - 'beer'
```

This results in

```
{
    'overlayfiles': [
        {'elements': ['some/path'], 'overlayname': 'root'},
        {'elements': ['abc', 'foo'], 'overlayname': 'ec2-1'},
        {'elements': ['bar', 'beer'], 'overlayname': 'ec2-2'}
    ]
}
```

The problem that 'root' is a key would be gone and we know about overlayname and its elements

schaefi:
and with namespaces

```yaml
overlayfiles:
  azure-common:
    -
      overlayname: 'root'
      elements:
        - common-feature1
        - common-feeature2

  azure-extra:
    -
      overlayname: azure-extra
      elements:
        - extra-stuff
```

results in

```
{
    'overlayfiles': {
        'azure-common': [
            {'elements': ['common-feature1', 'common-feeature2'], 'overlayname': 'root'}
        ],
        'azure-extra': [
            {'elements': ['extra-stuff'], 'overlayname': 'azure-extra'}
        ]
    }
}
```

Thoughts ?

schaefi:
But why are namespaces needed under overlayfiles ? Didn't we had a conversation about this part ? The profile information for which the namespace is used by convention is already present at a higher level in the document. I also thought we had an agreement on this that the only distinction we need to make within overlayfiles is the name of archives when they are used.

jgleissner:
keg merges the input data from all relevant directories into one dictionary. The namespaces are there to make sure parameters do not clash, unless you want to override data from an upper layer, in which case you use the same namespace. That's how it works for the profile configuration, the package lists, and the scripts (a.k.a. everything else) and I figured we use the same approach for the overlay files as well. If we switched to a flat list we lose the inheritance mechanism. I guess we could still merge lists but you couldn't overwrite parameters (not without inventing another mechanism). For example, in SLE15 SP3 in Azure, the overlay files for the emergency shell login changed, so the overlayfiles section in `data/csp/azure/sle15/sp3/overlayfiles.yaml` overwrite the key `azure-emergency-shell` to switch to the new implementation.

If you wanted to use flat lists you could simply add the overlay files configuration to the image definition, but that means if you for example add an Azure specific file you'd have to modify every image definition that uses the Azure module, and that's one thing keg is meant to avoid.

schaefi:
ok I understand your example. So namespaces are used in two different scopes:

* as profile names
* as names to allow for override

I'm fine with this design as long as using a namespace is an option and not mandatory. That means the code (templates in our case) needs to be able to differentiate that because a namespace when used changes the data type structure by an additional layer.

Regarding overlayfiles I was under the impression that it's possible to manage them without overrides.

To be honest the concept of allowed overrides with free form names makes it very very hard to understand the final results. An inheritance model as I thought about it was based on "base + more + more" and not "base + more - part-of-more + some-more".
I cannot imagine a jinja template that will be able to deal with each and every possible data setup this concept allows. From my perspective this is too flexible. I rather would allow a little bit more data repetition than trying the most flexible concept. But this is just my personal opinion and beyond the topic we discuss here.

jgleissner:
I've written some code to do a bit of testing with the updated keg-recipes to see whether things actually fall into place the way I intended, and I thought it might be useful to share it. It doesn't actually generate a tarball, but it processes the overlay configuration and generates an archive configuration. More specifically, it produces a new section 'archives' in the data dictionary, which contains the following structure:

```json
{
  "archives":
  {
    "tarball_name1":
    {
      "path_to_file_root1":
      [
        "etc/some_file",
        "usr/bin/another_one"
      ],
      "path_to_file_root2":
      [
        "foo/bar"
      ]
    }
  }
}
```

The idea is that the tarball generator only needs to iterate the archive section, bundle the files up, and drop them into the given file name. The code also adds an `archive` parameter to the profile sections for all archives != `root`, which would need to be evaluated in the schema. I've also added a cli parameter `--dump` which simply dumps the final data dictionary to stdout so you can actually see what happens, which I think should be useful for developing and debugging.

The code is in branch `overlay-test`, see commit https://github.com/SUSE-Enceladus/keg/commit/393259f9fbe781c41c3f2bd84874bbd85f95f1e4 (it's not very polished). Let me know what you think.

jesusbv:
When ok, I will squash the commits into one commit and fix the conflicts after.

kiwi_keg/generator.py
```diff
@@ -153,6 +155,94 @@ def create_custom_scripts(self, override: bool = False):
...
+    def _update_config_kiwi(self, archive_name, dest_dir):
+        if os.path.exists(dest_dir) and os.path.exists(self.kiwi_description):
+            etree_parser = etree.XMLParser(remove_blank_text=True)
+            kiwi_xml = etree.parse(self.kiwi_description, parser=etree_parser)
+            kiwi_root = kiwi_xml.getroot()
+
+            image_element = kiwi_root.find('.//packages[@type="image"]')
+
+            archive_element = etree.SubElement(image_element, 'archive')
```
rjschwei, line 236: But the archive element should not be created if we have `root.tar.gz`

kiwi_keg/keg.py
```diff
@@ -37,7 +38,11 @@
     -d DEST_DIR, --dest-dir=DEST_DIR
         Destination directory for generated description, default cwd
 
-    -l, --list-recipes
+    --tar-overlays
+        Option to create a tarball root.tar.gz in destination directory
+        [default: false]
```
jgleissner, line 43: Why would this be optional? Overlay is part of the image configuration, and as such the archives should be generated when keg generates the image description.

kiwi_keg/generator.py
jgleissner: Generating the kiwi file first without the archive tags and then modified it isn't so nice, and it also requires knowledge about its structure that is defined elsewhere. That is why I suggested a two-stage approach. First, process the overlay configuration info and add the appropriate tags to the dictionary that can be used for schema generation. Second, produce the tarballs. Did you have a look at my previous comment and the sample code?

I think copying the files into a temp dir might be unnecessary. It should be possible to loop over all files and add them to the tarball directly using `TarFile.AddFile`. The archive name of the file it can be set in the `TarInfo`, and at that opportunity the owner and group can be set to `root` as well, which we really should do anyway. Again, the sample code in https://github.com/SUSE-Enceladus/keg/commit/393259f9fbe781c41c3f2bd84874bbd85f95f1e4 preprocesses the overlay information in a way that would make it easy to do that.

jgleissner (changes requested):

kiwi_keg/keg.py
schaefi: After all the conversations we had, yes you are right the options needs to adapt. There is one situation which we need to take into account though. The "common" tarball can be either provided as "root.tar.gz" or as a directory named "root/...". I think the default for keg could be to create tarballs but there should be an option that allows to create the common root overlay as a directory structure. People who build with kiwi outside of obs never use a tarball for overlay files, they always use a directory structure which is much easier to manage than a binary blob.

jgleissner: I didn't think of unpackaged root overlay. But, the question is do we even want to support it? I don't see a lot of value in that. Working with files in a tarball is a PITA for sure, but obviously with keg you can work with your files directly in your recipes environment. If you're meddling with overlay files directly in your kiwi build setup, you're probably not a keg user.

schaefi: As I said there is a value. Almost nobody who builds images outside of obs uses tarballs. A keg user that builds the image description for use outside of obs will use overlay files outside of a binary tarball, that is something I know from all customers I had contact with in the last years. Also in yesterdays conversation the idea was shared that we provide the keg written image descriptions in some public space. Also for this use case you want to provide files and not binary data.
The entire requirement for overlay files as tarballs is because of the open buildservice which can only deal with source files.

> If you're meddling with overlay files directly in your kiwi build setup, you're probably not a keg user.

I doubt that, for the reasons mentioned above.

kiwi_keg/generator.py
jesusbv: > I think copying the files into a temp dir might be unnecessary. 

But it is not a temp dir, in an introductory conversation with @schaefi it was agreed the creation of `.tar.gz` files would be optional, and keg will always create an overlay tree structure (if `overlayfiles` is present) 

is this not the case ?

kiwi_keg/keg.py
jesusbv: I think a good option is to change the creation of the tarballs to be true by default and the flag to disable it

thoughts ?

kiwi_keg/generator.py
jesusbv: > But the archive element should not be created if we have `root.tar.gz`

I understood different, if root, do not add root to archive, AFAIU now is not even necessary.

jgleissner: Now I get it, the flag only declares whether the overlay should be tar'd or not, it will always be produced. So yes, if producing unpacked `root/` is an option we must have, then the option makes sense. I don't really think this is a work-flow that should be encouraged. You should update the recipe, run keg, then kiwi, not run keg, meddle with the overlay files, run kiwi, which would appear the prime reason to me why you wouldn't want it archived. But just my opinion.

However, this only works for root, doesn't it? I.e. named archives always need to be tarballs, right? The code doesn't create tarballs for overlays != root if the `tar_overlays` option is not set.

kiwi_keg/keg.py
schaefi: sounds good yes. As Joe mentioned it can only be supported for the files that would go to root.tar.gz. Those files which are configured as ```<archive name="..."/>``` has to be in an archive.

kiwi_keg/generator.py
jesusbv: > But the archive element should not be created if we have `root.tar.gz`

Moved the check for root inside the method.

jgleissner:
Where are we with this?

jesusbv:
Let's approve the PR before solving the conflicts,

Currently, by default tarballs are created, there is a flag to disable it (creating the overlay tree in that case)
Adds archive to xml if not root

rjschwei (approved):

kiwi_keg/generator.py
```diff
@@ -153,6 +156,94 @@ def create_custom_scripts(self, override: bool = False):
...
+
+                archive_element = etree.SubElement(image_element, 'archive')
+                archive_element.attrib['name'] = archive_name
+                tree = etree.ElementTree(kiwi_root)
+                tree.write(
+                    self.kiwi_description,
+                    encoding="utf-8",
+                    xml_declaration=True,
+                    pretty_print=True
+                )
```
jgleissner, line 245: Wouldn't this add all archives simply to the first `<packages type="image">` section? Doesn't really matter though, I still think we shouldn't handle it like that in the first place, i.e. rewrite the just generated `config.kiwi`. All the information is there as soon as keg parsed the input data, so there's no need to do it that way. It's needlessly complicated IMHO and it assumes a certain structure of the `config.kiwi` in keg which is actually defined in the schema and not by keg. As mentioned before, I'd rather preprocess the data in keg and generate the complete `config.kiwi` in one go.

jgleissner (changes requested):

kiwi_keg/generator.py
```diff
@@ -153,6 +156,94 @@ def create_custom_scripts(self, override: bool = False):
...
+                        tarball_data[overlay_name] = {}
+                        tarball_data[overlay_name] = overlay_dest_dir
+
+                    if not disable_tar_overlays:
+                        for overlay_name, overlay_dest_dir in tarball_data.items():
+                            overlay_tarball_name = '{}.tar.gz'.format(overlay_name)
+                            self._create_tarball(
+                                os.path.join(self.dest_dir, overlay_tarball_name),
+                                overlay_dest_dir
+                            )
```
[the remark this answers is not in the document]
jesusbv: Done

kiwi_keg/generator.py
jesusbv: True, and agreed, I rather doing that refactor once this is merged, in a new different PR

PR #36, opened 2021-03-10
Add option to list available recipes

### LINKED ISSUE #32: List available recipes
Following #17 conversation, 
it would be handy to know what recipes are available or what images can be created from the CLI

jesusbv:
This Fixes #32

### DIFF
--- a/kiwi_keg/image_definition.py
+++ b/kiwi_keg/image_definition.py
@@ -16,7 +16,6 @@
 # along with keg. If not, see <http://www.gnu.org/licenses/>
 #
 import os
-import yaml
 from typing import (
     List, Dict
 )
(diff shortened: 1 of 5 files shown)


schaefi (approved):
I like it :+1:

kiwi_keg/image_definition.py
```diff
@@ -136,3 +136,16 @@ def update_contents(self, include_paths):
...
+        images_files = KegUtils.get_all_files(self.image_root)
+
+        for image_file in images_files:
+            if os.path.basename(image_file) == 'image.yaml':
+                rel_path = os.path.relpath(
+                    os.path.dirname(image_file),
+                    self.image_root
+                )
+                images_recipes.append(rel_path)
+        return images_recipes
```
rjschwei, line 151: OK, our list returns something that looks like:

`images/sles-sap/ondemand/15-sp2`

but when I want to create an image description what is expected is

`keg -r foo -d bar sles-sap/ondemand/15-sp2`

at least according to the help, because it states:

`SOURCE    Path to image source, expected under RECIPES_ROOT/images`

That means a user that does a list cannot use the output directly in the image description creation command. That's not nice.

So we should either fix the listing output or we should modify the way `SOURCE` is processed.

jgleissner:
Sorry about the late comment. So this works with what's currently in `keg-recipes`, but it expects files called `image.yaml`. However keg doesn't really care about file names. It scans all directories in the given path from top to bottom for yaml files and merges the content into a dict, and expects to have a full image definition at that point. I've used file names to indicate that kind of content they specify for easier navigation, but for keg only the content is relevant. So if a user chooses a different naming scheme this implementation may miss image definitions.

PR #38, opened 2021-03-11
Fixed script templates

schaefi:
**script template file is not related to schema type**
    
The script template file names for config.sh.templ, images.sh.templ
were prefixed by the schema type, e.g 'vm'. This is not needed
because the contents of the scripts are defined by the collected
data from the image definition tree and do not change in ther
semantic depending on the schema type. Therefore we only need
one template for config.sh and one template for images.sh

**Fixed comment indentation in script templates**


jgleissner:
Fair point, but I still like the idea that people could use different templates adjusted for their needs. Plus we may want to use a simpler template for singlebuild.

schaefi:
not sure how the singlebuild template is connected with the scripts. Regarding the users choice there is nothing you would add into the template to change the structure of the script output. This is all done in the information you provide with the yaml. The script template just walks through a data structure and places raw data (with a small exception on services). Actually I would go even that far and handle the creation of the scripts as pure code in keg without any template engine around.

The template imho pays off for the creation of data with a semantic like the XML data for kiwi. At that point the template and also different ones per, we call it ```schema: name``` makes sense. For scripts I don't see the value of the template engine

This is probably something for another conversation.

Until then it would be good if we don't have to deal with custom templates for scripts and I hope you are fine with having one for each script we support. Please also have in mind with the former concept you are forced to provide a script template for each schema. At the moment this would be 'vm' and 'oem' and the templates for the script engine are the exact same files. It will not be any different for other schema as the main purpose is to handle different kiwi descriptions whereas all data for scripts is not connected to the semantic of the description.

I really hope for a thumps up on this as it also allows us to refactor the design for scripts at a later point in time

jgleissner:
Not using a template engine at all might be an option, especially if we conclude that there is no value in having the option of custom script templates. Wrt to singlebuild, I figured a simpler template would suffice because you don't need to have profile specific code section but then the multibuild template should handle that case fine, so I guess there's no real need for it. So let's merge this and revisit the topic later (if needed).

PR #40, opened 2021-03-16
Allow dict keys to be null when merging

jgleissner:
Allow keys to be 'null' when merging into data dictionary. This can be used to delete an inherited key.

This is used in https://github.com/SUSE-Enceladus/keg-recipes/pull/5.


schaefi:
@jgleissner I fear it's time to to write tests along with your code ;) The tests we wrote are good and valuable and even if you have negative experience with test driven development in the past I think it's a good time to come back to it as it is really worth the effort. tests are a nightmare when they are bad, but our tests here are good :)

test/unit/utils_test.py

is missing a test for your new added condition. Please add one.

Thanks

schaefi:
If you are unsure how to call the tests, we added some contributing info here

* https://github.com/SUSE-Enceladus/keg#contributing

or let's meet

rjschwei:
Is the `null` support intended to be permanent? I thought this temporary before we split the li/vli data out from azure.

jgleissner:
The delete key feature could generally be useful and is actually also uses in `data/csp/gce/sle15/sp2/overlayfiles.yaml` and `data/csp/ec2/sle12/sp5/overlayfiles.yaml`. I think it's also a better implementation than overwriting it with an empty key.

schaefi (approved):
Looks good :+1:

PR #41, opened 2021-03-17
Better performance for getting the files

jesusbv:


rjschwei:
I bet you cannot measure it ;) the better performance I mean. We'd probably have to have 1 million or more files in the tree for this to make a difference.

schaefi:
> I bet you cannot measure it ;) the better performance I mean. We'd probably have to have 1 million or more files in the tree for this to make a difference.

Jesus did measure it, which was the motivation for the change. We had a conversation prior the pull request. Also there is
https://www.python.org/dev/peps/pep-0471/

So this makes sense to me

schaefi (approved):
looks good to me :+1:

jgleissner:
I don't think this change is particularly useful. I can't imaging os.walk() will be a performance bottleneck with the number of files keg will typically have to deal with. Plus you're introducing recursion, which means the function may throw a maximum recursion exception if the directory tree to be scanned is deeper than the maximum recursion level python allows. Certainly a corner case, but still.

PR #43, opened 2021-04-08
Use template to add archive

jesusbv:


test/data/schemas/vm.kiwi.templ
```diff
@@ -83,16 +83,23 @@
...
     >
-        {%- for namespace, packages in content['packages'][pkg_type].items() %}
+	{%- if 'overlayfiles' in content.keys() %}
+	{%- for overlay, overlay_content in content['overlayfiles'].items() %}
+    	{%- if 'name' in overlay_content.keys() %}
+        <archive name="{{overlay_content['name']}}.tar.gz"/>
+    	{%- else %}
+	{%- if profile != 'common' %}
+        <archive name="{{profile}}.tar.gz"/>
+	{%- endif %}
```
jgleissner, line 93: As mention, this should produce duplicate `archive` tags in case there is more than one `overlayfiles` section in a profile. I think it's ok to have this in the test schema, for now anyway. I'll look into adapting the real schema file.

jgleissner (approved):
Code looks good to me.

rjschwei (approved):

jgleissner:
Can we merged this? @schaefi Any objections?

PR #44, opened 2021-04-20
Update documentation title

schaefi:
After a converstation with Stefan Knorr from the doc team
we agreed on updating the Keg title to be more specific.
In addition a sub-title was added to outline what keg is
good for

### DIFF
--- a/doc/xml/book.xml
+++ b/doc/xml/book.xml
@@ -2,11 +2,14 @@
       href="file:/usr/share/xml/docbook/schema/rng/5.1/docbookxi.rnc"
       type="application/relax-ng-compact-syntax"
      ?>
-<book xmlns="http://docbook.org/ns/docbook" xml:lang="en" version="5.1" xml:id="keg-documentation">
-  <title>Keg Documentation</title>
+<book xmlns="http://docbook.org/ns/docbook" xml:lang="en" version="5.1" xml:id="keg-reference-guide">
+  <title>Keg Reference Guide</title>
   <info/>
   <preface>
     <title>Preface</title>
+    <note>
+      <para>Template-Based KIWI Description Builder</para>
+    </note>
     <itemizedlist>
       <listitem>
         <para>
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -1,7 +1,10 @@
 .. keg documentation master file
 
-Keg Documentation
-=================
+Keg Reference Guide
+===================
+
+.. note::
+   Template-Based KIWI Description Builder
 
 .. toctree::
    :maxdepth: 1


doc/xml/book.xml
```diff
@@ -2,11 +2,14 @@
...
-<book xmlns="http://docbook.org/ns/docbook" xml:lang="en" version="5.1" xml:id="keg-documentation">
-  <title>Keg Documentation</title>
+<book xmlns="http://docbook.org/ns/docbook" xml:lang="en" version="5.1" xml:id="keg-reference-guide">
+  <title>Keg Reference Guide</title>
   <info/>
   <preface>
     <title>Preface</title>
+    <note>
+      <para>Template Based KIWI Description Builder</para>
+    </note>
```
sknorr, line 12: Wording sounds good! Only thing I would change about wording is add a `-`: `Template-Base KIWI Description Builder`.

However, structure-wise, this is not ideal: Is there a way you can make this come as `info/subtitle`? (I guess @tomschr might be able to answer that since he's created the script that converts to DocBook.)

sknorr (commented):
Looks good so far!

I'd suggest a structural change to a real subtitle however, so we actually get this as part of the output on navigation pages.

doc/xml/book.xml
sknorr: Update: I looked at https://docutils.sourceforge.io/FAQ.html#how-can-i-indicate-the-document-title-subtitle and that seems to indicate that the correct way to add a subtitle is this:

```
Keg Reference Guide
===================

Template-Based KIWI Description Builder
---------------------------------------
```

However, after being run through `make xml`, that supposed subtitle comes out as a new section (which is something that probably cannot be corrected by any tool thereafter because it looks just the same as a section that was intended to be there).
I guess it might make most sense to forget about specifically adding a `subtitle` to the document because it might turn out to be a waste of time, unless there's something obvious that I missed here.

schaefi: For that subtitle to work in both worlds (Sphinx) and the docbook conversion I believe we need Thomas Schraitle to change something in rstxml2docbook

sknorr: I looked at the output of `make xml` and I don't think Toms will be able to do anything about this, tbh, because what comes out looks exactly like a normal section. I can't imagine a heuristic that would correctly be able to fix this (unless this was a quirk behavior triggered by an own command-line switch like --first-section-title-is-subtitle).

tomschr: It seems, subtitles are not really supported. :disappointed:  As a quick note, I found here some information: https://github.com/sphinx-doc/sphinx/issues/3574 (I haven't read through it, but it doesn't look promising).

tomschr: @schaefi Stefan already summarized that quite good. I share his concerns. 

However, it seems, you need the subtitle only for the _documentation title_, not for every headers, right?

In that case, there might be a solution. After looking into the RST specification, I see these ideas:

1. Use [metadata](https://docutils.sourceforge.io/docs/ref/rst/directives.html#metadata)
2. Use a [role](https://docutils.sourceforge.io/docs/ref/rst/directives.html#role)

I would try the metadata first, something like this:

```rst
.. meta::
   :description: Keg Reference Guide
   :description: subtitle=Template-Based KIWI Description Builder
```

I haven't tried it so it's purely theoretical. :wink: If this data is transferred to the intermediate XML file, I could use that information to create the `subtitle` element.

tomschr: Okay, adding the above `meta` information to `source/index.rst` and running `make xml` gives me this:

```xml
<section ids="keg-documentation" names="keg\ documentation">
   <title>Keg Documentation</title>
   <meta content="Keg Reference Guide" name="description"></meta>
   <meta content="subtitle=Template-Based KIWI Description Builder" name="description"></meta>
```

The XSLT stylesheet could then look for a `<meta content>` structure.

Would that help?

schaefi: > However, it seems, you need the subtitle only for the documentation title, not for every headers, right?

yes only there

schaefi: > I would try the metadata first, something like this: ..meta ...

I have tried that. No errors but also no changes. Pretty interesting

tomschr: > > I would try the metadata first, something like this: ..meta ...
> 
> I have tried that. No errors but also no changes. Pretty interesting

That was expected. :wink: It's just meta information. My idea was to first come up with some markup solution and second how we could implement that. It seems, we need to circumvent that issue anyway.

If you think this `.. meta` could be helpful, we could think how to implement it to a) the HTML writer, and b) in the XSLT stylesheet for the RST -> DocBook conversation.

I never customized a writer, but it seems you need to derive the `HTMLTranslator` class from `/usr/lib/python3.?/site-packages/sphinx/writers/html.py`. The tricky part is to extract the information and add the right tag to the HTML output.

sknorr: To be honest, I didn't mean for this to spiral out in this way -- the subtitle suggestion was supposed to make the document more visible, and perhaps more of a nice-to-have. So, if this is going to involve significant amounts of coding, I will definitely not insist on there being a subtitle.

A very simple solution (which will however make the normal title very long) would be to just use this:

```
Keg Reference Guide: Template-Based KIWI Description Builder
============================================================
```

tomschr: That's an even better idea! :+1: I was under the impression this would be a "must-have" rather than a "nice-to-have". :wink:

schaefi: ok, I did not find a way to add a subtitle. That would really require to add meta info and Tom would need to adapt the rstxml2docbook. I think this is too much of an effort. With the current layout the information appears in the Preface and I'd say we live with that for the moment.

PR #46, opened 2021-04-22
Dump data dictionary

jgleissner:
Allows to dump the generated data dictionary on stdout instead of normal operation. Useful for debugging.

### DIFF
--- a/test/unit/generator_test.py
+++ b/test/unit/generator_test.py
@@ -157,7 +157,7 @@ class TestKegGenerator:
             dest_file['other'] = os.path.join(tmpdirname, 'other', 'etc', 'hosts')
             dest_file['other_aws'] = os.path.join(tmpdirname, 'other', 'etc', 'resolv.conf')
 
-            assert mock_shutil_copy.call_args_list == [
+            assert sorted(mock_shutil_copy.call_args_list) == sorted([
                 call('../data/data/overlayfiles/base/etc/hosts', dest_file.get('root_base')),
                 call('../data/data/overlayfiles/csp/aws/etc/resolv.conf', dest_file.get('root_csp_aws')),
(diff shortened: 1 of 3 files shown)


test/unit/generator_test.py
```diff
@@ -157,7 +157,7 @@ def test_create_overlays(
             dest_file['other'] = os.path.join(tmpdirname, 'other', 'etc', 'hosts')
             dest_file['other_aws'] = os.path.join(tmpdirname, 'other', 'etc', 'resolv.conf')
 
-            assert mock_shutil_copy.call_args_list == [
+            assert sorted(mock_shutil_copy.call_args_list) == sorted([
```
jesusbv, line 160: does this have any effect on `dump-dict` debugging option ?

jgleissner: Not directly, but unit test would fail locally for me without this change. The order of calls to `shutil.copy()` is not set in stone, as the `os.scandir()` does not return directory entries in any particular order. Sorting the args lists before comparing solves this. I should have probably put it into a separate PR.

jesusbv: > Not directly, but unit test would fail locally for me without this change. The order of calls to `shutil.copy()` is not set in stone, as the `os.scandir()` does not return directory entries in any particular order. Sorting the args lists before comparing solves this. 

Yes, good idea

> I should have probably put it into a separate PR.

That sounds good to me

jgleissner: https://github.com/SUSE-Enceladus/keg/pull/47 includes the `dump-dict` feature (even though I didn't really intend it). However, that PR changes those tests and so the PRs would be incompatible anyway, so I'm just closing this one.

PR #47, opened 2021-04-23
Update overlay archive support

jgleissner:
This PR is related to https://github.com/SUSE-Enceladus/keg-recipes/pull/12.

First of all, sorry this PR turned out to be way bigger than anticipated. The reason is piecing the overlay data together in the schema template is rather fiddly and the templates are messy enough as-is. So I moved the bit that figures out which archives need to be created from the function that creates the overlay archives into the image definition. This makes things easier in the template. This information can also be used in `create_overlays` (no reason to figure out the archives again), and hence the function could be simplified.

Here are the details of the changes:

- Processing of overlay information

When the image definition is populated, the `overlayfiles` sections are evaluated and the archive names and contents are derived from that. The archive names are injected into the `profiles` sections (or `content` section for single build) of the data dictionary for use in schema template. The archive names and the directories to be included in them are injected into the data dictionary for use with archive creation.

- Archive creation

Since the archives and the desired contents are already available in the data dictionary, the `create_overlays` function simply uses that information to produce the tarballs. Tarballs are now created directly without intermediate copy, and file ownership is set to root. In case `--disable-root-tar` is set, files earmarked for that archive are copied into `root` instead of a tarball being produced.

- Mockup recipes data

The mockup recipes data in `test/data` wasn't quite structured in a way `keg` expected the data, and since I had to change some of it to make it work with the modified code base, I updated it to be in line with the structure of the actual `keg-recipes` data.


kiwi_keg/keg.py
```diff
@@ -39,10 +40,14 @@
...
-        Option to disable the creation of a tarball root.tar.gz in destination directory.
-        If present, an overlay tree would be created instead.
+        Option to disable the creation of root.tar.gz in destination directory.
+        If present, an overlay tree will be created instead.
         [default: false]
 
+    --dump-dict
+        Dump generated data dictionary to stdout instead of generating an image
+        description. Useful for debugging.
+
```
rjschwei, line 50: It is probably useful to have the generated image definition and the debug output, I think there is no reason to have them mutually exclusive. Of course one can run keg and generate the image definition and the run keg again in debug mode, if that is the intention I am OK with that.

rjschwei (approved):

kiwi_keg/keg.py
jgleissner: I implemented it this way because I wanted it to dump the dictionary to stdout rather than logging it as a debug message, so you could e.g. feed it to another tool, and I felt that makes it more of a different operating mode. I used it mostly to verify that the data structures where as intended while working on the code, and didn't really care about the generation part at that point. If people prefer that it can do both things in one go I can change it. I don't expect it to be very useful outside development though.

kiwi_keg/generator.py
```diff
@@ -155,73 +156,63 @@ def create_custom_scripts(self, override: bool = False):
...
+                        ) -> None:
         """
-        Copy all the files and the overlay tree structure from overlays section under root inside destination directory.
-
-        :param: bool tarball:
-            Flag to create tarball
+        Create overlay archives as defines in the 'archives' section of the
+        data dictionay.
+
+        :param: bool diasable_root_tar:
```
jesusbv, line 168: s/diasable/disable

kiwi_keg/generator.py
```diff
@@ -155,73 +156,63 @@ def create_custom_scripts(self, override: bool = False):
...
+                        overwrite: bool = False,
+                        compression: str = 'gz',
+                        ) -> None:
         """
-        Copy all the files and the overlay tree structure from overlays section under root inside destination directory.
-
-        :param: bool tarball:
-            Flag to create tarball
+        Create overlay archives as defines in the 'archives' section of the
+        data dictionay.
```
jesusbv, line 166: s/dictionay/dictionary

kiwi_keg/generator.py
```diff
@@ -155,73 +156,63 @@ def create_custom_scripts(self, override: bool = False):
...
+                        disable_root_tar: bool = False,
+                        overwrite: bool = False,
+                        compression: str = 'gz',
+                        ) -> None:
         """
-        Copy all the files and the overlay tree structure from overlays section under root inside destination directory.
-
-        :param: bool tarball:
-            Flag to create tarball
+        Create overlay archives as defines in the 'archives' section of the
```
jesusbv, line 165: s/defines/defined ?

jesusbv (approved):
LGTM

PR #49, opened 2021-05-05
Properly support overwriting of keys

jgleissner:
Keys from upper levels may be overwritten in lower levels, but adding subsequent data modules may revert that by merging upper level dict files again. This commit prevents duplicate merging.


rjschwei (approved):
LGTM

Issue #39, opened 2021-03-11
create scripts without a template engine


schaefi:
If you look at the templates config.sh.templ and images.sh.templ you will see that they almost completely just traverse a dictionary and print the data. There is no semantic or syntax that the template would enforce. Thus creating the same file with just python would mean reading of the dict and writing its values into a file.

From that perspective the creation of the script code could be done simpler and without a template engine in between

jgleissner:
Forgot to close this, sorry.

PR #50, opened 2021-05-14
Generate config.sh and images.sh directly

jgleissner:
This moves generating the config scripts into keg rather than relying on templates in keg-recipes. Use of templates is still supported to generate script headers which can be used to provide additional information available from the data dictionary.

This PR also includes some changes to drop any single build 'special' sauce. It wasn't really practical to separate those two things.


jgleissner:
Related issue: https://github.com/SUSE-Enceladus/keg/issues/39

jesusbv (approved):
LGTM

jgleissner:
I renamed `utils.py` to `file_utils.py` because I didn't really like having `utils` and `script_utils`. I dropped the additional namespace-only class from `file_utils` as I didn't see the point.

rjschwei (approved):

PR #52, opened 2021-07-02
Add documentation on keg recipes data strucutre

jgleissner:
This PR adds sections `Recipes Basics`, `Image Definition`, and `Data Modules` which describe the fundamental data structure keg uses and how it merges data. The `Data Modules` part may be a bit hard to read, not sure. I thought it might be useful to have all specifications in one place rather than broken into several bits, but it means you'll probably have to scroll to see the explanation below. Feedback welcome.

PR #54, opened 2021-07-08
Fix missing mypy library stubs

schaefi:
For static type checking many modules startet to create
library stubs packages to contain the types, like header
files in C/C++. This commit installs the missing stubs
package to the development environment


schaefi:
This will also make the tests to work in #52

rjschwei (approved):
LGTM

PR #60, opened 2021-08-25
Nested profiles

jgleissner:
Add support for nested profiles in multi-build image definitions.
This potentially allows for shorter package sections, less duplication in overlay archives, and shorter config.sh scripts.
This commit also streamlines the produced overlay archives by not including directories multiple times.


jgleissner:
Can we merge this? @schaefi Any objections?

PR #65, opened 2021-09-23
Added keg obs service

schaefi:
The fetch_from_keg obs service can be used to integrate
the creation of image descriptions via keg with the open
build service. This commit creates the basic structure
for the service.


schaefi:
sorry wrong branch

PR #66, opened 2021-09-23
Added keg obs service

schaefi:
The fetch_from_keg obs service can be used to integrate
the creation of image descriptions via keg with the open
build service. This commit creates the basic structure
for the service.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,6 @@
 # Manifest describing source package contents
 
-graft kiwi_keg/tasks
+graft kiwi_keg/obs_service
 graft package
 
 include Makefile
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,11 @@ install:
 	for man in doc/build/man/*.8.gz; do \
 		install -m 644 $$man ${buildroot}usr/share/man/man8 ;\
 	done
+	# keg obs service
+	install -d -m 755 ${buildroot}usr/lib/obs/service
+	mv ${buildroot}usr/bin/fetch_from_keg ${buildroot}usr/lib/obs/service
+	install -m 644 kiwi_keg/obs_service/fetch_from_keg.service \
+		${buildroot}usr/lib/obs/service
 
 tox:
 	tox "-n 5"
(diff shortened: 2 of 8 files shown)


schaefi:
[keg & the build service](https://trello.com/c/2GBPbynM/1792-keg-the-build-service)

schaefi:
_service example

```xml
<services>
    <service name="fetch_from_keg">
        <param name="git-recipes">https://github.com/SUSE-Enceladus/keg-recipes.git</param>
        <param name="image-source">leap/jeos/15.2</param>
    </service>
</services>
```

schaefi:
Please note this is a start, the implementation to fill the service with live needs to follow based on the review of the code so far. Thanks

jesusbv (approved):
LGTM

rjschwei:
I like `fetch_from_keg` as it implies there is a pint in my near future ;) . That said, I think `compose_kiwi_description` would be more explanatory to those that are not familiar with keg. LGTM generally

schaefi:
> I like `fetch_from_keg` as it implies there is a pint in my near future ;) . That said, I think `compose_kiwi_description` would be more explanatory to those that are not familiar with keg. LGTM generally

Ohh really, we have such nice names for our projects and now we start to make them boring but explanatory :-)) Nah just kidding. I'm fine with a rename if the head architect @jgleissner agrees with it too

jgleissner:
Looks good, and I really like `fetch_from_keg` too, but I agree that a more descriptive service name might be better.

One thing to bear in mind perhaps, we'll need support for multiple repositories since we also have a private repo, and we'll need to be able to handle a deploy token for it, that we may not want to store in `_service` file.

jgleissner (approved):

schaefi:
I found out that the source service automatically adds something like ```--outdir /home/ms/Project/sbs/external/home:sax2/test-keg-service/tmp5nc4627w.compose_kiwi_description.service``` to the service command call. This means the program called has to understand this option

schaefi:
I also setup a package build using the service:

* https://build.opensuse.org/package/show/home:sax2/test-keg-service

As you can see it does not work because on the remote OBS end our service does not exist. This is expected but we have to have in mind when we are done we have to ask the OBS team to install our service

PR #74, opened 2021-10-28
Complete obs service

### LINKED ISSUE #68: Implement obs service
With #66 the basic layout of the service was created but so far it does nothing.
This issue exists to track the actual implementation of running keg as part of the obs-service

schaefi:
This commit implements the actual keg call and store
of result image data in the output directory from
which OBS auto commits the data to the project if
called on the remote side. This Fixes #68

### DIFF
--- a/kiwi_keg/obs_service/compose_kiwi_description.py
+++ b/kiwi_keg/obs_service/compose_kiwi_description.py
@@ -46,22 +46,50 @@ Options:
         At the time the service is called through the OBS API
         this option is set.
 """
+import os
 import docopt
-import logging
 
+from kiwi.utils.temporary import Temporary
+from kiwi.command import Command
+from kiwi.path import Path
 from kiwi_keg.version import __version__
-
-log = logging.getLogger('keg')
-log.setLevel(logging.INFO)
+from kiwi_keg.image_definition import KegImageDefinition
+from kiwi_keg.generator import KegGenerator
 
 
 def main() -> None:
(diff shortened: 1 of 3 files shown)


rjschwei (approved):

rjschwei:
> The service IMO should simply read the version number from the existing `config.kiwi` (if there is one) and increment the patch level by 1. For new images we'd need a default, 1.0.0 or maybe 0.9.0 or something. The maintainer can always adjust it if needed.
> 

We have documented the version convention. While in development image for the given distro in development get 0.9.x. For FCS we start out with 1.0.0

> The other thing that we probably want to do at this point is generating a changelog, but that should probably be a separate PR. It's not terribly straight-forward to generate the changelog. We could capture the commit hash so the next service run can use this as a starting point. The trickier question is how to figure out the relevant commits. Some sort of manual tagging in the commit message is probably not the way to go. Maybe keg could keep track of all source files that were used to generate the requested image description, and hand that over to the service so it can use that information to gather the appropriate commit messages. Thoughts?

There is a changelog generator that is part of the go service, we should look at what was done there.

schaefi:
> There is a changelog generator that is part of the go service, we should look at what was done there.

I usually create changelogs from git history. The same could be done in this service very easily because the git is already checked out. I have code in python to create this changelog information but if you guys prefer to do this in go this is also fine by me :) In any case it should be discussed and agreed in an extra issue imho. I haven't see that this topic was created and once this PR is merged the information about changelogs is more or less lost in space

PR #76, opened 2021-11-24
Better spacing in config.sh

jgleissner:
Do not append newlines at the end of sections.
Insert newline between if blocks.
Insert newline between scriptlets.


rjschwei (approved):

jgleissner:
pytest is failing because `test/data/keg_output_xml/config.xml` doesn't match the generated output (anymore). This file was already changed in commit 69a09f720458aee0f454701bea8d95e0c8f795a0 to adjust for a change of indentation step width in kiwi. It appears this has changed once again. Maybe we should use `xmldiff` to compare the files so at least the test won't fall over anymore over changes in whitespaces.

jgleissner:
I'll fix this in a separate PR, as it's unrelated to the changes in this one.

PR #77, opened 2021-11-24
Use xmldiff to compare generated config.xml to expected

jgleissner:
Use xmldiff to compare generated config.xml to expected version instead of filecmp to avoid failing test due to immaterial differences in indentation.


rjschwei (approved):

schaefi (approved):
nice :+1:

PR #78, opened 2021-11-24
Support for generating source info logs

jgleissner:
Add an option to generated source information logs. When used, keg writes a file `log_sources` (for single-build) or multiple files
`log_sources_{profile}` (for multi-build). This information can be used to automatically generate a change log from git commit messages that contains all messages applying to the given profile, but only those.

The source log format is:
```
range:start:end:file
pathspec
```
The first form is for ranges, i.e. parts of source files.
The second form is for files or directories included in full, i.e. config snippet files and overlay directories.

When the option is used, `keg` will use the `SafeTrackerLoader` implementation in `file_utils.py` which generates extra tags to store in the dictionary when parsing yaml. Instead of a regular dict, `AnnotatedMapping` will be used. This basically works like a normal dictionary, but hides the extra tags when accessed through standard interfaces. Functions that need to be aware of the tags check the dictionary type and use special methods to get the extra info.

I have also written a script that generates change logs from source log files. I'll create a separate PR for that if this one is accepted.


jgleissner:
The fix for the failing test is in https://github.com/SUSE-Enceladus/keg/pull/77.

rjschwei (approved):

PR #79, opened 2022-01-13
Add changelog generator

jgleissner:
Add a script that generaters a changelog (text and yaml format supported) from
the git commit history and the source info log produced by keg.

### DIFF
--- a/kiwi_keg/changelog_generator/generate_recipes_changelog.py
+++ b/kiwi_keg/changelog_generator/generate_recipes_changelog.py
@@ -78,7 +78,7 @@ def get_commits_from_path(pathspec, rev=None):
 
 
 def get_commits(gitargs):
-    sp = subprocess.run(args=gitargs, capture_output=True)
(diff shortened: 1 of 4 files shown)


jgleissner:
Unit tests are broken. They work locally so this may need some time for debugging.

rjschwei (approved):

jgleissner:
I see where the problem with the unit tests is. I use real data from the repository's commit log, but in the CILint the commit history isn't available so this fails. Seems I'll have to feed it fake data instead. I guess that's the better option anyway.

jgleissner:
Unit tests are fixed, but it appears the `capture_output` parameter I use with subprocess isn't supported by python 3.6. Compatibility with python 3.6 may not be a bad idea, so I'll look into changing it.

jgleissner:
This should be ok for merge now, another look-over may be not a bad idea. I just noticed an issue with the implementation, namely that it doesn't work with multiple data roots. This should be fixable by storing the data path separately in the source log. I will create a separate PR for that.

jesusbv (approved):
LGTM

rjschwei (approved):

PR #85, opened 2022-02-15
fix: include top keys in source info tracking

jgleissner:
Include the source info for top level keys when generating source logs.

### DIFF
--- a/kiwi_keg/source_info_generator.py
+++ b/kiwi_keg/source_info_generator.py
@@ -57,6 +57,8 @@ class SourceInfoGenerator:
             with self._open_source_info_file('log_sources', overwrite) as outf:
                 for r in self.image_definition.recipes_roots:
                     outf.write('root:{}\n'.format(r))
+                outf.write(self._get_source_info_top())
+                outf.write('\n')
                 outf.write(self._get_source_info_profile('common'))
                 outf.write('\n')
         else:
@@ -65,6 +67,7 @@ class SourceInfoGenerator:
                     continue
                 if profiles[profile_name].get('base_profile'):
                     continue
+                top_src_info = self._get_source_info_top()
                 base_src_info = self._get_source_info_profile(profile_name)
                 nested_profiles = profiles[profile_name].get('nested_profiles')
                 if nested_profiles:
@@ -74,6 +77,8 @@ class SourceInfoGenerator:
                         ) as outf:
                             for r in self.image_definition.recipes_roots:
                                 outf.write('root:{}\n'.format(r))
+                            outf.write(top_src_info)
+                            outf.write('\n')
                             outf.write(base_src_info)
                             outf.write('\n')
                             outf.write(self._get_source_info_profile(nested_profile))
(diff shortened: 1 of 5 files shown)


rjschwei (approved):

kiwi_keg/source_info_generator.py
```diff
@@ -112,18 +124,18 @@ def _get_source_info_profile(self, profile_name):
 
     def _get_mapping_sources(self, mapping):
```
jesusbv, line 125: should this method be renamed ?
before it skipped `nested_profiles` and `base_profile` keys now, there is a `skip_keys` to not handle,
and given the implementation of ` _get_source_info_top` is the top source info comes from this method, right ?

maybe `_get_top_mapping_sources` ?

jgleissner: The functionality of `_get_mapping_sources` hasn't changed. It returns the source info of the given mapping and if any key in it is also a mapping it traverses it too. It is now called on the root of the dictionary but still also individually on the profile nodes, as those end up in different source logs. At some keys it need to stop traversing, otherwise every log would contain all source info which isn't wanted.

However, implanting that knowledge in an otherwise rather generic function is perhaps not very nice. It's an internal function so no real biggie I guess, but it would be nicer if the info about keys it should skip would be handed over to the function by the caller, which has the knowledge on what node it calls `_get_mapping_sources`. I'll change accordingly.

jgleissner: I fiddled a bit with the code but it didn't really get prettier. ;) All of the keys in `skip_keys`, with the exception of `profiles`, is just in there to suppress a warning. Not sure whether moving this around is worth it. I could actually downgrade the warning to a debug message (as it really is more of a debug thing) and the drop the key list. Or just leave it as it is.

jesusbv: > I fiddled a bit with the code but it didn't really get prettier. ;) 

So, my question about the rename was because this is a fix to include top keys, then there is a method 
that get top keys which calls `get_mapping_sources` and return the result. 
I thought the top info selection/inclusion would happen in `_get_source_info_top` which I understand is a cleaner
way to do it (following the same as `_get_source_info_profile`) and saw the changes in `get_mapping_source`
and that's why I thought if that method should be renamed, as if the skip keys list is _always_ skipped 
then it should be renamed, if it is not, then the idea you proposed of sending said list from the caller sounds good to me

> All of the keys in `skip_keys`, with the exception of `profiles`, is just in there to suppress a warning. Not sure whether moving this around is worth it. I could actually downgrade the warning to a debug message (as it really is more of a debug thing) and the drop the key list. Or just leave it as it is.

OK, AFAIU now `get_mapping_sources` always return top keys and the skipping list was introduced to avoid a warning message.

jgleissner: Ok, I renamed `_get_source_info_top` to `_get_source_info_root` to make it clearer that it traverses the dict from its root, rather than just returning info for top keys. I also remove `profiles` from the `skip_keys` and rather pass it as an option from `_get_source_info_root`.

PR #86, opened 2022-02-18
compose_kiwi_description: abort on empty change logs

jgleissner:
If the image does not have any changes (i.e. generated change logs for all flavors are empty), abort and delete all generated files.

The current implementation only detects if the repositories don't have new commits, but since most commits in `keg-recipes` are likely to only affect a subset of all images, this won't be enough to prevent unnecessary version bumps and rebuilds. This PR adds support to handle this situation properly.

### DIFF
--- a/kiwi_keg/obs_service/compose_kiwi_description.py
+++ b/kiwi_keg/obs_service/compose_kiwi_description.py
@@ -59,7 +59,6 @@ import docopt
 import itertools
 import logging
 import os
-import shutil
 import sys
 
 from kiwi.xml_description import XMLDescription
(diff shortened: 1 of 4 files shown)


rjschwei (approved):

kiwi_keg/obs_service/compose_kiwi_description.py
```diff
@@ -252,9 +262,23 @@ def main() -> None:
                 )
             image_version = get_image_version(os.path.join(args['--outdir'], 'config.kiwi'))
 
-        for source_log in glob.glob(os.path.join(args['--outdir'], 'log_sources*')):
-            flavor = source_log[len(os.path.join(args['--outdir'], 'log_sources')) + 1:]
-            generate_changelog(source_log, args['--outdir'], flavor, image_version, rev_args)
+        for source_log, flavor in get_log_sources(os.path.join(args['--outdir'])):
+            changes_filename = f'{flavor}{"." if flavor else ""}changes.yaml'
+            changes_path = os.path.join(args['--outdir'], changes_filename)
+            have_changes |= generate_changelog(source_log, changes_path, image_version, rev_args)
```
jesusbv, line 268: I don't understand that bit, so `generate_changelog` returns `return result.returncode == 0` (line 170)
which means the method/command run inside the method did not produce any errors, right ?
My question is: if no errors == there are changes in the changelog ?

jgleissner: Basically yes. `generate_recipes_changelog` returns 2 in case there wasn't an error but no new changes either. The `generate_changelog` functions abort on error and otherwise translates the exist status to True/False.

jesusbv: Ack, that makes sense to me now
Maybe a comment to clarify that ?

jesusbv (approved):
LGTM

kiwi_keg/obs_service/compose_kiwi_description.py
jgleissner: Not a bad idea. :)

PR #90, opened 2022-03-02
Use rst link syntax instead of markdown

dcermak:


jgleissner (approved):
I hadn't even noticed that the links were misformatted. Thanks!

PR #96, opened 2022-05-18
Codename "minikeg"

jgleissner:
This PR substantially changes the structure of the input data used with keg. It removes the concept of keg profiles which keg used to generate `profiles` and `preferences` sections and instead requires the input data to form those sections. The input data's structure is now aligned with the structure that `config.kiwi` uses. Instead of using a jinja2 template for producing a `config.kiwi` file (thought still supported as an option), keg now has a built-in XML generator. See https://github.com/SUSE-Enceladus/keg/blob/minikeg/notes_minikeg.rst for details about the design changes.

This PR also adds a image definition schema which keg will use to verify the structure of the input data. This should help with producing more helpful error message in case there are structural errors in the recipes data. It also makes some bits of the code simpler as extra checks whether mandatory keys exists are not necessary after the validation step. It doesn't check whether parameter values are valid (i.e. accepted by kiwi), as full validation is still performed using kiwi after the image description has been generated.

I created the PR against a staging branch since the keg-recipes data needs to be converted before it is usable with this update and that way we can refine this code stream and its equivalent on keg-recipes side until happy and then merge at the same time.

There is a `minikeg` branch in https://github.com/SUSE-Enceladus/keg/tree/minikeg which has a couple of new image definitions in https://github.com/SUSE-Enceladus/keg-recipes/tree/minikeg/images/minikeg but since most data in the repo is still old style it may be confusing for getting an idea how the new data structure looks like. The unit tests have simpler image definitions that should be a lot easier to read, see https://github.com/SUSE-Enceladus/keg/tree/minikeg/test/data.


kiwi_keg/dict_utils.py
```diff
@@ -0,0 +1,72 @@
+# Copyright (c) 2021 SUSE Software Solutions Germany GmbH. All rights reserved.
```
rjschwei, line 1: 2022 on my calendar

kiwi_keg/generator.py
```diff
@@ -103,6 +110,32 @@ def create_kiwi_description(self, overwrite: bool = False) -> None:
             kiwi_config.write(kiwi_document)
             kiwi_config.write('\n')
 
+    def create_xml_description(self, overwrite) -> None:
```
rjschwei, line 113: Not doing anything with `overwrite` argument. Presumably there should be a check

``if not overwrite and os.path.exists(....):
      raise .....
```

**Review by rjschwei — CHANGES_REQUESTED** [review]

**Comment by rjschwei** [comment]
IF we want to keep a rational and history of design changes etc. we should probably have a `doc/design` dir. In thta case what's in the .rst file would be appropriate in that directory. Otherwise many components should be in the README file.

**Review by rjschwei — APPROVED** [review]

File: kiwi_keg/keg.py [hunk 3]
```diff
@@ -139,14 +143,19 @@ def main():
...
         )
-        image_generator = KegGenerator(
-            image_definition=image_definition,
-            dest_dir=args['--dest-dir']
-        )
         if args['--dump-dict']:
+            try:
+                image_definition.populate()
+            except KegError:  # pragma: no cover
+                pass
```
jesusbv, line 150: dumping the dictionary is not important ?
as in, if it fails we don't care ?

kiwi_keg/generator.py
```diff
@@ -43,7 +47,7 @@ class KegGenerator:
     :param str dest_dir: Destination directory
     """
     def __init__(
-        self, image_definition: KegImageDefinition, dest_dir: str
+            self, image_definition: KegImageDefinition, dest_dir: str, archs: list = []
```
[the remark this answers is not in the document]
jgleissner: Thanks, fixed.

kiwi_keg/keg.py
jgleissner: `image_definition.populate()` may raise due to the new schema validation, but we still want to be able to dump the dictionary for debugging purposes anyway, so the error is ignored at this point.

PR #97, opened 2022-05-24
Obs service convert params

jgleissner:
Convert flag parameters 'disable-version-bump', --disable-changelog-update', '--disable-revisions-update', and '--force' to string type parameters. OBS src service cannot handle flags.
Add parameter '--image-version' to set image version directly.


jgleissner:
Unfortunately a couple of unit tests fail because kiwi has changed upstream and now the expected and the actual schema version are different. I've fixed this in the `minikeg` branch already, so may not be worth it fixing it again in this branch.

rjschwei (approved):

jesusbv (approved):
LGTM

PR #111, opened 2022-11-11
Release keg 2.0.0

jgleissner:
This cleans up a few things and bumps the version to 2.0.0.

I've also added a spec file that can be used directly, which will be useful for converting the package to an OBS src service one. I also included the intersphinx mapping file so man pages can be built offline. Also a requirement for building this as a src service.


package/python-kiwi_keg.spec
```diff
@@ -0,0 +1,138 @@
...
+%global pygroup python
+%global sysgroup admin
+%global develsuffix dev
+%else
+%global pygroup Development/Languages/Python
+%global sysgroup System/Management
+%global develsuffix devel
+%endif
+
+Name:           python-kiwi_keg
```
rjschwei, line 45: Should be `python3-kiwi-keg` since we are not building a Python 2 version. While the package provides an executable and an API and it would be appropriately named `python-kiwi-keg` the precedence we have set is that we use `python3-` as prefix when no Python2 version is being built. Also means the file name of the spec file needs to change.

package/python-kiwi_keg.spec
```diff
@@ -0,0 +1,138 @@
...
+Version:        2.0.0
```
rjschwei, line 46: There was no API change from version 1.1.0 that I remember, as such this should be `1.2.0`. I reviewed the changes and think everything we merged since 1.1.0 falls into the "functionality in a backwards compatible manner" category. Of course I could have missed something.

rjschwei (changes requested):

package/python-kiwi_keg.spec
jgleissner: Almost no package in Factory uses `python3` as prefix for the package name. `python` appears to be the standard (including python packages that have python2 build disabled). The binary packages are automatically produced with the appropriate prefix, i.e. `python38`, `python39`, and `python310` in Factory and `python3` in SLE/Leap.

PR #116, opened 2022-11-18
Rename rpm

jgleissner:


package/python-kiwi-keg.spec
```diff
@@ -84,14 +84,15 @@ Requires:       python%{python3_pkgversion}-Cerberus
 %else
 Requires:       python%{python3_pkgversion}-cerberus
 %endif
+Obsoletes:      python%{python3_pkgversion}-kiwi_keg <= 2.0.1
```
rjschwei, line 87: We have not released the package to any distro, I would drop the `Obsoletes:` directive, it is only useful for updates. If we have the `Obsoletes` directive we also need to have a `Provides:` directive see https://en.opensuse.org/openSUSE:Package_dependencies#Renaming_a_package . I'd drop this stuff as it is related to updating and there is nothing to update as we have not really released the package.

rjschwei (changes requested):
See comment about Obsoletes directive

package/python-kiwi-keg.spec
jgleissner: Fair point, removed.

rjschwei (approved):

PR #119, opened 2023-01-12
Fix unit tests of OBS src service module

jgleissner:
Fix a previously non functional assertion.
Remove a previously non functional assertion for a function call that was not to be executed in that circumstance anyway.

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -5,21 +5,26 @@ on: [push]
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.6, 3.8]
+        python-version: [3.8, "3.10"]
+        os: [ubuntu-latest]
+        include:
+          # Use ubuntu 20.04 for python 3.6 (dropped from later versions)
+          - python-version: 3.6
+            os: ubuntu-20.04
 
     steps:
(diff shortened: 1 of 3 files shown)


jgleissner:
CI tests for python 3.6 are broken at the moment (python 3.6 not supported anymore in used platform), looking into it.

rjschwei:
@jgleissner @apozsuse recently fixed the Python 3.6 unit test issues for other projects we can take the same solution.

rjschwei (approved):
This also works for me, we just have to be careful not to introduce python 3.6 incompatible syntax as we will no longer catch it during testing.

apozsuse:
> @jgleissner @apozsuse recently fixed the Python 3.6 unit test issues for other projects we can take the same solution.

It was actually Sean that fixed it for some repo (and I extended to other repos having the same issues).

The problem is that python 3.6 is not available any longer in ubuntu:latest container (the one we use for the ci/cd pipeline), and we just splitted the test for 3.6 in a separate job running on ubuntu:20.04.

[Here](https://github.com/SUSE-Enceladus/ec2imgutils/pull/99/files) you can find a PR with that change.

jgleissner:
> The problem is that python 3.6 is not available any longer in ubuntu:latest container (the one we use for the ci/cd pipeline), and we just splitted the test for 3.6 in a separate job running on ubuntu:20.04.

That's what I did too, just a little differently.

PR #142, opened 2023-07-20
compose_kiwi_description: fix JSON serialization error

jgleissner:
On recent python version, when converting an existing YAML change log to a JSON one, an exception can occur because when parsing YAML date strings may be created as datetime objects, which json.dump() cannot serialize. Set 'default=str' in json.dump() to auto-convert those to strings.


jesusbv (approved):
LGTM

PR #147, opened 2023-09-08
Add source service commands

smarlowucf:
Add the commands to install the source service and run the service with osc.

### DIFF
--- a/doc/source/obs_source_service.rst
+++ b/doc/source/obs_source_service.rst
@@ -67,8 +67,22 @@ updated. Default is `true`.
 If true, refresh image description even if there are no new commits. Default
 is `false`.
 
-The system the source service is run on needs to have `keg` and
-`obs-service-keg` installed. Refer to the `Using Source Services
+The system the source service is run on needs to have the
+`compose_kiwi_description` service installed. Use the following command to
+install the service in openSUSE:
+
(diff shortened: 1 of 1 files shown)


smarlowucf:
Not sure how best to include these commands. Mentioning osc without previous notes could be confusing to someone who doesn't have osc installed. And explicitly mentioning openSUSE install command leaves out other os's. But maybe there's an assumption if you are in the OBS section that you already have knowledge of OBS/OSC?

rjschwei (approved):

jgleissner:
> Not sure how best to include these commands. Mentioning osc without previous notes could be confusing to someone who doesn't have osc installed. And explicitly mentioning openSUSE install command leaves out other os's. But maybe there's an assumption if you are in the OBS section that you already have knowledge of OBS/OSC?

We could add a link to https://en.opensuse.org/openSUSE:OSC so people have an easy way to look for information if they are not familiar with osc and how to get the package. The target audience is certainly people who are familiar with OBS though.

jgleissner (approved):

smarlowucf:
Added a comment directing to the osc docs.

jgleissner:
> Added a comment directing to the osc docs.

Looks good, thanks!

Issue #115, opened 2022-11-18
Do not generate config.sh when there are no changes


rjschwei:
When there are no changes to the scriptlets keg still generates config.sh. The file gets a new generated time stamp in the file, but the change is superfluous, we should not generate the file if there are no material changes.

jgleissner:
This isn't quite as straight-forward as it sounds. Keg simply generates files based on input. It doesn't know about changes per-se. The `compose_kiwi_description` script used with the OBS source service has some idea of changes between image descriptions, but it's limited to extracting the image version from `config.kiwi` and checking whether the source repo has new commits and whether the `generate_changelog` command produced new entries. It doesn't have any deeper knowledge about the generated content. It would certainly be possible to implement a special check for config.sh diffing only the part below the header to catch this case, and then delete only the newly generated config.sh to avoid OBS copying over the old one. That seems a bit hacky to me, especially considering that the sole purpose is just to avoid a new time stamp in the file.

We could simply drop the time stamp from the config.sh header. The file's time stamp would still change but it wouldn't result in a diff in OBS.

rjschwei:
I am OK with dropping the time stamp. It may be advantageous to drop the information from config.kiwi as well and write a separate file that contains the information about the latest image description generation. While theoretically we would bump the version every time the image gets generated and as such the timestamp in config.kiwi does not generate a "diff because of timestamp" separating the generation time information from the image description content is probably a good idea in general.

jgleissner:
Implementation merged.

Issue #107, opened 2022-09-28
Change log generator misses some changes


jgleissner:
Keg source tracking stores the source information of each key in the internal dictionary which can be used to generated a change log from the commit info. Not all changes are caught though. If a key was changed from e.g.:

```yaml
some_key:
  - foo
  - bar
  - baz
```
to
```yaml
some_key:
  - foo
  - bar
```
the source info generated by keg for `some_key` would be line 1 to 3 for whatever file `some_key` was imported from. The only line that changed was line 4 though, so querying the git commit history for changes in line 1-3 would not return the commit, and hence no change log entry is generated for that change.

gmoro:
This is an important bug to have fixed as it is a bit disrupting for the workflow we are envisaging for ALP and Minimal-VM.
As a work around we can use "force" in the OBS service to regenerate the kiwi files, but this still will live the changelog with the wrong entries, and this will impact the future feature that we had planned of generating .changes files from the keg changelogs.

jgleissner:
There shouldn't be any wrong entries, just missing ones. Since the composed image description may have many different source fragments from different source files, keg tracks from which source files and lines keys where imported that end up in the final dictionary and writes them to a file. The change log generator uses that information to generate the log (or logs specific to profiles in case of multi-build). Simply using the commit message of all involved files may include unrelated commits, hence the source tracking by keg.

However, as stated, there are some cases that slip through, i.e. when a key value was truncated as in the above sample, or removed entirely. Keg cannot know from looking at the current state that there used to be a key in a previous state. You could work around this by replacing removed trailing lines in keys with empty lines instead of completely cutting them. Removed keys could be replaced by `some_key: Null` instead of cutting them. It's not great, admittedly.

Fixing this for good is tricky. Essentially we would need to check out the previous version from the recipes repo, build the image dictionary with keg and dump it somehow, then generate the image dictionary of the current version and look for deletions and somehow figure out which commits removed them. This would make change log generation significantly more complicated than it already is. I'll try to look into this some time, but I don't see a quick solution.

jgleissner:
[...]

PR #148, opened 2023-10-26
compose_kiwi_description: check for changed files

jgleissner:
Check generated files for changes compared to existing ones and delete unchanged files.

Add option to check existing image description for stale files (i.e. files not generated by the new image definition), and remove those (enabled by default).

Do not abort anymore if the generated change log has no new entries, but base decision to abort or proceed on result of the above checks.

This addresses https://github.com/SUSE-Enceladus/keg/issues/107 and https://github.com/SUSE-Enceladus/keg/issues/115.


kiwi_keg/tools/compose_kiwi_description.py
```diff
@@ -86,21 +88,32 @@
         converted if necessary. Conversion from 'osc' is not supported.
         [default: json]
 
+    --purge-stale-files=<true|false>
```
jesusbv, line 91: could we use the presence of the flag as true and the lack of presence/omission as false instead of setting a value ?
i.e.
`cmd ... --purse-stale-files` -> True, purge

jesusbv: If the default behaviour is to purge, maybe same approach of present or not present but different name
`--not-purge-stale-files` if not present, files would be purged

jesusbv (approved):
other than the meaning/use of one of the flags, LGTM

rjschwei (approved):

PR #150, opened 2024-06-06
Fix timer disable

jgleissner:
Also fix a typo.


jgleissner:
Investigating unit tests.

rjschwei (approved):

PR #151, opened 2024-06-07
Build with newer Python interpreter

rjschwei:
The code requires capture_output as an argument to subprocess which was introduced in Python 3.7, thus we can no longer builds against Python 3.6 default on SLE 15 and openSUSE Leap 15 and have to build against a newer version of Python available with 15.4 and later.


jgleissner (changes requested):

rjschwei:
Superseded by #152

PR #152, opened 2024-07-03
Support newer python versions

schaefi:
This patch is many fold

1. Update spec file to build against py311 on SLE/Leap

2. Fix tox setup to work with newer python versions

3. Fix sdist tarball layout, for proper submissions to pypi and also according to standard the tarball layout has to match the name of the module. The name of the module is kiwi_keg not keg

4. Don't use source services for packages maintained in SLE, they are not allowed there anyway. Fix the Makefile build target such that a simple "make build" creates all files for proper package submissions

5. Fix github actions workflow to run properly against all tox targets and python versions we want to support, 3.6,3.11,312

You can find package builds available here:

* https://build.opensuse.org/project/show/home:marcus.schaefer:branches:OBS_Maintained:python-kiwi-keg

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,8 @@ build: clean tox
 	sed -ie "s@>=[0-9.]*'@'@g" setup.py
 	# build the sdist source tarball
 	$(python) -m build
+	mv dist/kiwi_keg-${version}.tar.gz dist/keg-${version}.tar.gz
+	rm -f dist/*.whl
 	# restore original setup.py backed up from sed
 	mv setup.pye setup.py
 	# update rpm changelog using reference file
@@ -63,10 +65,8 @@ build: clean tox
 	# copy spec file
 	cp package/$(PKG).spec dist/
 
-pypi: clean tox
-	$(python) setup.py sdist upload
-
 clean: clean_git_attributes
 	$(python) setup.py clean
 	rm -rf doc/build
 	rm -rf doc/dist
+	rm -rf dist
--- a/package/python-kiwi-keg.spec
+++ b/package/python-kiwi-keg.spec
@@ -16,14 +16,13 @@
 # Please submit bugfixes or comments via https://bugs.opensuse.org/
 #
 
-
 %if 0%{?suse_version} > 1500
 %bcond_without libalternatives
 %else
 %bcond_with libalternatives
 %endif
 
-%define         skip_python2 1
+%{?sle15_python_module_pythons}
 Name:           python-kiwi-keg
 Version:        2.1.1
 Release:        0
@@ -32,32 +31,41 @@ Summary:        KEG - Image Composition Tool
 License:        GPL-3.0-or-later
 Source:         keg-%{version}.tar.gz
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
-BuildRequires:  %{python_module Jinja2}
-BuildRequires:  %{python_module Sphinx}
-BuildRequires:  %{python_module base >= 3.6}
 BuildRequires:  %{python_module setuptools}
-BuildRequires:  fdupes
-BuildArch:      noarch
-Requires:       python-Jinja2
-Requires:       python-PyYAML
-Requires:       python-docopt
-Requires:       python-schema
-Requires:       python3-kiwi >= 9.21.21
-%if %python_version_nodots < 37
-Requires:       python-iso8601
+BuildRequires:  %{python_module Sphinx}
+BuildRequires:  %{python_module Jinja2}
+BuildRequires:  %{python_module schema}
+BuildRequires:  %{python_module pytest}
+BuildRequires:  %{python_module docopt}
+BuildRequires:  %{python_module PyYAML}
+# On TW the kiwi package is always the latest python default: python3-kiwi
+# On Leap <15.5 and SLES15 <SP5 only one python exists: python3-kiwi
+%if 0%{?suse_version} > 1600 || 0%{?sle_version} < 150500
+BuildRequires:  python3-kiwi
+%else
+BuildRequires:  %{python_module kiwi}
 %endif
+BuildRequires:  fdupes
 %if %{with libalternatives}
-Requires:       alts
 BuildRequires:  alts
+Requires:       alts
 %else
 Requires(post): update-alternatives
 Requires(postun):update-alternatives
 %endif
-%if "%{python_flavor}" == "python3" || "%{python_provides}" == "python3"
+Requires:       python-Sphinx
+Requires:       python-Jinja2
+Requires:       python-schema
+Requires:       python-docopt
+Requires:       python-PyYAML
+%if 0%{?suse_version} > 1600 || 0%{?sle_version} < 150500
+Requires:       python3-kiwi
+%else
+Requires:       python-kiwi
+%endif
 Provides:       python3-kiwi-keg = %version
 Obsoletes:      python3-kiwi-keg < %version
-%endif
-
+BuildArch:      noarch
 %python_subpackages
 
 %description
@@ -77,7 +85,7 @@ more given git repositories that contain keg-recipes source tree. It supports
 auto-generation of change log files from commit history.
 
 %prep
-%setup -q -n keg-%{version}
+%setup -q -n kiwi_keg-%{version}
 
 %build
 # Build Python 3 version
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.8, 3.9, "3.10", 3.11]
+        python-version: ["3.11", "3.12"]
         os: [ubuntu-latest]
         include:
           # Use ubuntu 20.04 for python 3.6 (dropped from later versions)
@@ -16,21 +16,16 @@ jobs:
             os: ubuntu-20.04
 
     steps:
-      - uses: actions/checkout@v2
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - name: Python${{ matrix.python-version }}
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install
+      - name: Install Tox
         run: |
           python -m pip install --upgrade pip
           python -m pip install tox
-      - name: Check code
-        run: |
-          tox -e check
-      - name: Run unit tests
-        run: |
-          tox -e unit
-      - name: Test documentation build
-        run: |
-          tox -e doc
+      - name: Run linter, unit tests and render docs
+        run: tox -e "doc,check,unit_py${PY_VER/./_}"
+        env:
+          PY_VER: ${{ matrix.python-version }}
--- a/setup.cfg
+++ b/setup.cfg
@@ -17,3 +17,7 @@ ignore = E501
 [doc8]
 max-line-length = 90
 verbose = 1
+
+[mypy]
+ignore_missing_imports = True
+no_implicit_optional = False
--- a/tox.ini
+++ b/tox.ini
@@ -10,62 +10,89 @@
 #    $ tox -e doc
 #
 [tox]
+minversion = 3.3.0
+isolated_build = True
 skip_missing_interpreters = True
 skipsdist = True
 envlist =
     check,
-    unit_py36,
-    unit_py37,
-    unit_py38,
-    unit_py39,
-    unit_py310,
-    unit_py311,
-    unit_py312,
+    unit_py3_6,
+    unit_py3_11,
+    unit_py3_12,
     doc
 
 [testenv]
-allowlist_externals = *
+description =
+    {unit_py3_6,unit_py3_11,unit_py3_12}: Unit Test run with basepython set to {basepython}
+    check: Linter Checks
+    doc: Render Docs
+allowlist_externals =
+    bash
+    travis-sphinx
+    cp
+    make
+    rm
+    mv
+    flake8
+    python
+    pytest
 basepython =
-    unit_py36: {env:TOXPYTHON:python3.6}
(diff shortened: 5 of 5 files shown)


schaefi:
This PR would make #151 obsolete given it can be merged in the future. I also think #151 should not be merged for the same dependency issues explained here. For a path forward we should first make sure all python modules exists on target

package/python-kiwi-keg.spec
```diff
@@ -32,32 +31,39 @@ Summary:        KEG - Image Composition Tool
...
-Requires:       python3-kiwi >= 9.21.21
-%if %python_version_nodots < 37
-Requires:       python-iso8601
+BuildRequires:  %{python_module Sphinx}
+BuildRequires:  %{python_module Jinja2}
+BuildRequires:  %{python_module schema}
+BuildRequires:  %{python_module pytest}
+BuildRequires:  %{python_module docopt}
+BuildRequires:  %{python_module PyYAML}
+%if 0%{?suse_version} > 1500
```
rjschwei, line 41: Shouldn't this be `< 15400`. Python 3.11 is only available in SLE 15 SP4 and later. In cases where build in the build service is enabled for 15 SP4 or earlier we'd fall back to `python3-kiwi` For everything else, i.e. Factory and SLE 15 Sp4 and later we'd pick up the expansion from the `python_module` macro.

package/python-kiwi-keg.spec
```diff
@@ -32,32 +31,39 @@ Summary:        KEG - Image Composition Tool
...
 Requires(post): update-alternatives
 Requires(postun):update-alternatives
 %endif
-%if "%{python_flavor}" == "python3" || "%{python_provides}" == "python3"
+Requires:       %{python_module Sphinx}
+Requires:       %{python_module Jinja2}
+Requires:       %{python_module schema}
+Requires:       %{python_module docopt}
+Requires:       %{python_module PyYAML}
+%if 0%{?suse_version} > 1500
```
rjschwei, line 59: Ditto

schaefi:
Regarding python311-docopt I was wrong, it's already there in

```
<bdep name="python311-docopt" version="0.6.2" release="150600.16.3" arch="noarch" hdrmd5="104e6dfb3cd999a60ec136ab0668d0ea" project="SUSE:SLE-15-SP6:GA" repository="standard"/>
```

schaefi:
The python-kiwi update for SLES is prepared here:

* https://github.com/SUSE/kiwi_sle15/pull/12

package/python-kiwi-keg.spec
schaefi: good point, I added the suggested version constraint. Please also note that we have python3-kiwi for the default python in TW and changing that will require an upstream conversation. That will not be so easy because that .spec file is also used by other distributors. It might be needed that we add a suse specific patch in this case. Because of that I still kept the "> 1500" version constraint

schaefi: I made a kiwi fix to address this:

* https://github.com/OSInside/kiwi/pull/2581

schaefi: Once the kiwi in SLE is out and the kiwi in TW we don't need special sauce here.

Makefile
```diff
@@ -1,6 +1,6 @@
 buildroot = /
 docdir = /usr/share/doc/packages
-python_version = 3
+python_version = 3.11
```
jgleissner, line 3: Does hardcoding 3.11 here make sense? Wouldn't that break when running under python 3.12?

schaefi: Ah good catch, no doesn't make sense and was only a temporary change while I was testing the package. Fixed it.

schaefi:
Task list for me as a reminder

- [ ] submit python-schema to SLES15
- [ ] submit python-kiwi to SLES15 
- [ ] submit python-kiwi to Factory

If you can think of more, please add to the list above. Thanks

package/python-kiwi-keg.spec
schaefi: and here is the patch for kiwi upstream regarding the versioned provides from someone who really knows how it should be done :) https://github.com/OSInside/kiwi/pull/2583

schaefi:
* SR for python-kiwi to Factory: 1185374

schaefi:
@jgleissner @rjschwei forget what I said about SLES. I thought keg is something that we have in SLES but that's wrong, we have keg in the openSUSE area provided via Cloud:Tools. All dependencies are satisfied there, so we should be good.

schaefi:
@jgleissner correct me when wrong 

```
osc maintained python-kiwi-keg

openSUSE:Backports:SLE-15-SP5:Update/python-kiwi-keg
openSUSE:Backports:SLE-15-SP6:Update/python-kiwi-keg
```

So I created an `osc mbranch python-kiwi-keg` and will build the package there and then run a submit request, once we agree we are good to go here. 

yes/no/maybe ? Thanks

schaefi:
submission branch

* https://build.opensuse.org/project/show/home:marcus.schaefer:branches:OBS_Maintained:python-kiwi-keg

schaefi:
@jgleissner I added eventually missing dependencies (e.g python-schema on SLE) to Cloud:Tools such that keg becomes installable by just adding the Cloud:Tools project repo. Do I have your permission to update python-kiwi-keg to build for >= 311 in Cloud:Tools too ? Thanks

package/python-kiwi-keg.spec
```diff
@@ -77,7 +85,7 @@ more given git repositories that contain keg-recipes source tree. It supports
 auto-generation of change log files from commit history.
 
 %prep
-%setup -q -n keg-%{version}
+%setup -q -n kiwi_keg-%{version}
```
jgleissner, line 88: Any specific reason for that change? This breaks when using the OBS source service. I used `keg` because that's what the `obs_scm` service creates when checking out the keg repository, and I couldn't find an easy way to change that directory name.

package/python-kiwi-keg.spec
```diff
@@ -32,32 +31,41 @@ Summary:        KEG - Image Composition Tool
...
 %else
 Requires(post): update-alternatives
 Requires(postun):update-alternatives
 %endif
-%if "%{python_flavor}" == "python3" || "%{python_provides}" == "python3"
+Requires:       %{python_module Sphinx}
+Requires:       %{python_module Jinja2}
+Requires:       %{python_module schema}
+Requires:       %{python_module docopt}
+Requires:       %{python_module PyYAML}
```
jgleissner, line 60: This should be `Requires: python-xxx` etc. not `Requires: %{python_module xxx}`, otherwise it creates a dependency to all available versions.

package/python-kiwi-keg.spec
```diff
@@ -32,32 +31,41 @@ Summary:        KEG - Image Composition Tool
...
+%if 0%{?suse_version} > 1600 || 0%{?sle_version} < 150500
+Requires:       python3-kiwi
+%else
+Requires:       %{python_module kiwi}
```
jgleissner, line 64: Same here, should simply be `python-kiwi`.

jgleissner (changes requested):

package/python-kiwi-keg.spec
schaefi: Yes I had a specific reason and also wrote it in the commit message ;) ```keg``` is not the name of the python module. The name of the python module is ```kiwi_keg``` and the standard way to produce the dist target with the python build module automatically creates the source tarball with the correct names. As keg is also present on pypi this policy and actual requirement under certain conditions should not be violated.

I see you are using the recompress service in the _service definition and that creates you the tarball. But all of this is outside any standards and keg is also on pypi which should be considered the upstream. There is no correlation what the buildservice does here and all of this tar_scm stuff is a nightmare to maintain. Yes it makes your life easy as it might reduce explicit submission to obs but it hurts if you look at keg as a real opensource project with an upstream dist tarball on pypi, nothing matches than anymore

Regarding the obs service to build the package. I was pretty much confused what we are doing here. As said the source tarball does not follow python standards and the package build is a mix of data from ```_service``` plus some static files. The changes file does not correlate with the git history and the overall idea to build the package files is now a mix of scripts.

In addition no submission to the SUSE namespace will accept any ```_service``` file. Probably this is not a topic because Alex told me that we never wanted to have keg in SLES.

And the last part I'd like to mention, maybe I have forfeit my influence on keg when I left, but the idea for package maintenance on the keg project was as simple as this:

For upstream maintenance

```
bumpversion patch|minor|major
git tag --push
```

* github action triggers the pypi upload

For OBS maintenance

```
make build
cp dist/* /to/some/OBS/target
osc sr
```

* produces the exact same source tarball than upstream
* handles changes such that they are not out of sync with the git history
* is compatible with any submission target

I can imagine none of you is really interested in this sort of changes which I can understand as there is so much other stuff to do. But I strongly recommend to at least drop keg from pypi if the current package maintenance model should stay. If there is interest to maintain a healthy upstream, things needs to change.

just my2cents

package/python-kiwi-keg.spec
schaefi: yes sorry my bad, cut&paste bug. Thanks

package/python-kiwi-keg.spec
schaefi: also fixed

package/python-kiwi-keg.spec
schaefi: To make my point more clear. 

* .github/workflows/ci-publish-to-pypi.yml

```
- name: Prepare
      run: |
        python -m build
    - name: Publish
      uses: pypa/gh-action-pypi-publish@master
```

This is correct and the desired way to publish a python module to pypi

and in now obs

* Cloud:Tools/python-kiwi-keg

```
<services>
    <service name="obs_scm" mode="manual">
        <param name="filename">keg</param>

...

<service name="recompress" mode="manual">
    <param name="file">*.tar</param>
    <param name="compression">gz</param>
```
 
This is not the same and disconnected to upstream

jgleissner: > Yes I had a specific reason and also wrote it in the commit message ;) `keg` is not the name of the python module. The name of the python module is `kiwi_keg` and the standard way to produce the dist target with the python build module automatically creates the source tarball with the correct names. As keg is also present on pypi this policy and actual requirement under certain conditions should not be violated.

Well, I never liked `kiwi_keg` to start with, because the k in `keg` already stands for `kiwi`. ;)

The package is named `python-kiwi-keg` and thus complies with the naming convention. I don't really consider the name of the automatically generated tarball which is only used in the build service subject to that naming convention, but I guess it's debatable.

> I see you are using the recompress service in the _service definition and that creates you the tarball. But all of this is outside any standards and keg is also on pypi which should be considered the upstream. There is no correlation what the buildservice does here and all of this tar_scm stuff is a nightmare to maintain. Yes it makes your life easy as it might reduce explicit submission to obs but it hurts if you look at keg as a real opensource project with an upstream dist tarball on pypi, nothing matches than anymore

Using `obs_scm` and `recompress` seems pretty common to me in the build service. I don't consider pypi to be upstream. The github repo is the source of truth and pypi is a distribution, in a way similar to what OBS produces.

> Regarding the obs service to build the package. I was pretty much confused what we are doing here. As said the source tarball does not follow python standards and the package build is a mix of data from `_service` plus some static files. The changes file does not correlate with the git history and the overall idea to build the package files is now a mix of scripts.

I don't follow. Everything is generated from source including the change log.

> In addition no submission to the SUSE namespace will accept any `_service` file. Probably this is not a topic because Alex told me that we never wanted to have keg in SLES.

That's not correct. Server-side service must not be used in in maintained IBS projects.  Manual services are fine and there are `_service` files in IBS.

> And the last part I'd like to mention, maybe I have forfeit my influence on keg when I left, but the idea for package maintenance on the keg project was as simple as this:

You didn't forfeit anything. :)

> For upstream maintenance
> 
> ```
> bumpversion patch|minor|major
> git tag --push
> ```
> 
> * github action triggers the pypi upload
> 
> For OBS maintenance
> 
> ```
> make build
> cp dist/* /to/some/OBS/target
> osc sr
> ```
> 
> * produces the exact same source tarball than upstream

Debatable. :) It'll happily pick up local changes. Generally I'm not a fan of packaging up what's on a developers machine. Directly generating from source makes sure it's clean.

> * handles changes such that they are not out of sync with the git history
> * is compatible with any submission target
> 
> I can imagine none of you is really interested in this sort of changes which I can understand as there is so much other stuff to do. On the other side I believe upstream sources and what suse builds should be the same

As said, I consider the git repo to be the upstream source and what the src service generates is equivalent to the source tarball you can download from github. Said that, I don't really care. If having the same source tarball in OBS as in pypi is important for some reason we can switch methods.

schaefi: > Well, I never liked kiwi_keg to start with, because the k in keg already stands for kiwi

It was named that way because "keg" was already taken on pypi. 

> The package is named python-kiwi-keg and thus complies with the naming convention. 

This is not true and a typical "SUSE we know better answer" sorry to say so but the upstream for kiwi-keg is pypi not github. People who would package it for other distros would take it from pypi because it releases there with a version. The namespace for keg is kiwi_keg as you can see it in any of your imports and a tarball for a python project should be created using python build utilities and not some "SUSE recompress service"

> I don't consider pypi to be upstream

That says everything, so why do we have it there ?

schaefi:
I should have clarified this earlier, sorry my fault. There is a different mindset in packaging python modules. Both have there advantages/disadvantages but this PR will not lead to a conclusion in the current form and I should not have started anything without a conversation beforehand. Thus closing

jgleissner:
> I should have clarified this earlier, sorry my fault. There is a different mindset in packaging python modules. Both have there advantages/disadvantages but this PR will not lead to a conclusion in the current form and I should not have started anything without a conversation beforehand. Thus closing

Hang on, let's not overreact. The work you did on this PR is useful, needed, and appreciated. AFAICS the only disputed issue is the tarball name, which really boils down to using OBS source service or not.

rjschwei:
I think it makes sense to build keg with a newer version of Python. That said I also think it makes sense to try and be as generic as possible such that we do not have to make major spec file changes every time there is a change in the interpreter.

PR #155, opened 2025-04-22
chore(spec): convert to pyproject macros and prepare for Python 3.14.…

gubasso:
… Implement `https://build.opensuse.org/request/show/1241670` changes.

### DIFF
--- a/package/python-kiwi-keg.spec
+++ b/package/python-kiwi-keg.spec
@@ -1,7 +1,7 @@
 #
 # spec file for package python-kiwi-keg
 #
-# Copyright (c) 2023 SUSE LLC
+# Copyright (c) 2025 SUSE LLC
 # Copyright (c) 2022 SUSE Software Solutions Germany GmbH, Nuernberg, Germany.
 #
 # All modifications and additions to the file contributed by third parties
@@ -31,11 +31,16 @@ URL:            https://github.com/SUSE-Enceladus/keg
 Summary:        KEG - Image Composition Tool
 License:        GPL-3.0-or-later
 Source:         keg-%{version}.tar.gz
+# PATCH-FIX-UPSTREAM https://github.com/SUSE-Enceladus/keg/pull/154 Fix link to OBS documentation
+Patch0:         fix-doc-link.patch
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 BuildRequires:  %{python_module Jinja2}
 BuildRequires:  %{python_module Sphinx}
 BuildRequires:  %{python_module base >= 3.6}
+BuildRequires:  %{python_module pip}
 BuildRequires:  %{python_module setuptools}
+BuildRequires:  %{python_module sphinx_rtd_theme}
+BuildRequires:  %{python_module wheel}
 BuildRequires:  fdupes
 BuildArch:      noarch
 Requires:       python-Jinja2
@@ -51,7 +56,7 @@ Requires:       alts
 BuildRequires:  alts
 %else
 Requires(post): update-alternatives
-Requires(postun):update-alternatives
+Requires(postun): update-alternatives
 %endif
 %if "%{python_flavor}" == "python3" || "%{python_provides}" == "python3"
 Provides:       python3-kiwi-keg = %version
@@ -77,17 +82,17 @@ more given git repositories that contain keg-recipes source tree. It supports
 auto-generation of change log files from commit history.
 
 %prep
-%setup -q -n keg-%{version}
+%autosetup -p1 -n keg-%{version}
 
 %build
 # Build Python 3 version
-%python_build
+%pyproject_wheel
 
 # Build man pages
 make -C doc man
 
 %install
-%python_install
+%pyproject_install
 make buildroot=%{buildroot}/ docdir=%{_defaultdocdir}/ install
 
 %python_clone -a %{buildroot}%{_bindir}/keg


rjschwei:
We want to switch this over to a single build as well, i.e. define `pythons` etc. see any of the metadata packages as an example. Drop the `alternatives` etc.

package/python-kiwi-keg.spec
```diff
@@ -31,11 +31,16 @@ URL:            https://github.com/SUSE-Enceladus/keg
 Summary:        KEG - Image Composition Tool
 License:        GPL-3.0-or-later
 Source:         keg-%{version}.tar.gz
+# PATCH-FIX-UPSTREAM https://github.com/SUSE-Enceladus/keg/pull/154 Fix link to OBS documentation
+Patch0:         fix-doc-link.patch
```
jgleissner, line 35: No patch required. The doc fixes have already been merged.

jgleissner (changes requested):

rjschwei:
> > We want to switch this over to a single build as well, i.e. define `pythons` etc. see any of the metadata packages as an example. Drop the `alternatives` etc.
> 
> I'd prefer this to be multi python since it is more of an openSUSE package, but then kiwi (which keg requires) is set up as single python and unless @schaefi can be convinced otherwise, there's no point in a multi-python keg build.

I do not see the benefit of having a multi-python package. While we have modules there is no known use case where we have a user that **must** have Python version X, where X is different from the primary interpreter in TW. For Leap the multi-python idea does not really exist. Interpreters that get added to Leap generally do not have module support. In 15 only the 3.6 and 3.11 Python interpreters have modules are the other interpreters in that distro are bare bones, i.e. the Interpreter itself + pip and a couple of other modules. So this makes keeping builds working more difficult. I think if we focus on the primary interpreter we'll have less hassle.

package/python-kiwi-keg.spec
gubasso: > No patch required. The doc fixes have already been merged.

Done!

gubasso:
I am working on the spec file changes to support a single python3 build.

rjschwei:
> I updated the spec file and tested the build here: https://build.opensuse.org/package/show/home:gbasso:branches:Cloud:Tools/python-kiwi-keg
> 
> It is failing for some targets. For example, on Tumbleweed (https://build.opensuse.org/package/live_build_log/home:gbasso:branches:Cloud:Tools/python-kiwi-keg/openSUSE_Tumbleweed/x86_64):
> 
> ```
> [   15s] + /usr/bin/python3.11 -mpip wheel --verbose --progress-bar off --disable-pip-version-check --use-pep517 --no-build-isolation --no-deps --wheel-dir ./build .
> [   15s] /var/tmp/rpm-tmp.vrvfWs: line 55: /usr/bin/python3.11: No such file or directory
> [   15s] error: Bad exit status from /var/tmp/rpm-tmp.vrvfWs (%build)
> ```
> 
> Is it expected @jgleissner @rjschwei ? Or what am I missing for this spec file implementation (https://build.opensuse.org/projects/home:gbasso:branches:Cloud:Tools/packages/python-kiwi-keg/files/python-kiwi-keg.spec?expand=1) ?

The primary Python in TW is not python3.11 so this should not be building.

rjschwei:
And looking at the changed spec file I see you did not take the previous comment about looking at the metadata packages and how to setup a single python build up under consideration.

I really mean it, if you setup the spec file like https://github.com/SUSE-Enceladus/gcemetadata/blob/master/python-gcemetadata.spec it will work and it will make our lives easier.

jgleissner:
@gubasso, with single python build we won't need the `%python_clone` and `%python_expand` macros, and anything with `alternative` can be dropped too.

gubasso:
@jgleissner , @rjschwei ... I could build it correctly for almost all versions, except for `SLE_15_SP2` and `SLE_12_SP5`.

https://build.opensuse.org/package/show/home:gbasso:branches:Cloud:Tools/python-kiwi-keg

For those versions, I tried adding some conditionals in the spec file to make the build paly nice with the restrictions that showed up. But turns out there are quite a few incompatibilities that probably need to be handled elsewhere in the project (though I cloud be wrong).

One example: `SLE_15_SP2` and `SLE_12_SP5` couldn't resolve the `sphinx` dependencies used for building the docs. I tried making that part optional with some if/else statements and while that fixed the docs issue, it just led to another error popping up.

jgleissner:
We don't really care about old SLES version for this build, so the workaround for 15 SP3 can be dropped. We won't need  `%define skip_python2 1` either. And `%define skip_fdupes 1` can also be dropped and `%python_expand fdupes ...` changed to `%fdupes ...`.

I'm also a bit confused about those provides:
```
Provides:      python310-kiwi-keg = %{version}
Provides:      python311-kiwi-keg = %{version}
Provides:      python312-kiwi-keg = %{version}
Provides:      python313-kiwi-keg = %{version}
```
That doesn't feel correct, as we are only providing %{pythons}-kiwi-keg. Is there a reason for this?

package/python-kiwi-keg.spec
```diff
@@ -12,62 +12,72 @@
...
+# ------------------------------------------------------------------
+# SLE-15-SP3 ships python-setuptools 40.x – too old for a plain
+# PEP-517 build.  Everywhere else we have >= 42.
+#
+#   new_setuptools = 1   … modern setuptools (default)
+#   new_setuptools = 0   … SLE-15-SP3 or older
+# ------------------------------------------------------------------
+%define new_setuptools 1
+%if 0%{?suse_version} <= 150300
+%define new_setuptools 0
```
rjschwei, line 34: I do not really see a reason when we need to build this tool for anything older than 15 SP4. Or expressed differently older than Python 3.11 which is defined at the beginning with:
`%{?sle15_python_module_pythons}` as such this can be dropped.

package/python-kiwi-keg.spec
```diff
@@ -12,62 +12,72 @@
...
 %endif
 
-%python_subpackages
+%define skip_python2 1
```
rjschwei, line 37: Not needed

package/python-kiwi-keg.spec
```diff
@@ -12,62 +12,72 @@
...
+BuildRequires: %{pythons}-Sphinx
+BuildRequires: %{pythons}-sphinx_rtd_theme
+BuildRequires: python-rpm-macros
+BuildRequires: fdupes
+BuildRequires: make
+Provides:      python3-kiwi-keg = %{version}
+Provides:      python310-kiwi-keg = %{version}
+Provides:      python311-kiwi-keg = %{version}
+Provides:      python312-kiwi-keg = %{version}
+Provides:      python313-kiwi-keg = %{version}
```
rjschwei, line 61: These provides are not correct, we only want the `Obsoletes`

package/python-kiwi-keg.spec
```diff
@@ -77,41 +87,49 @@ more given git repositories that contain keg-recipes source tree. It supports
 auto-generation of change log files from commit history.
 
 %prep
-%setup -q -n keg-%{version}
+%autosetup -p1 -n %{upstream_name}-%{version}
+
+%if !0%{?new_setuptools}
```
rjschwei, line 92: Not needed

rjschwei (changes requested):

jgleissner (approved):
LGTM, thanks!

